PR #23, opened 2020-11-05
Load firewalld rules immediately

julmon:
Ansible `firewalld`module permits to load the new rule immediately, meaning that if used we don't have to reload `firewalld` manually to apply the new rules.


vibhorkumar123:
Thank you!!

PR #34, opened 2020-11-15
Improve PgBouncer databases and users management

jt-edb:
The `lineinfile` module provides more fine grained lines management than the "replace the whole file or nothing" approach provided by `template` module usage.


vibhorkumar123:
Merged the changes proposed by @jt-edb

PR #40, opened 2020-11-17
Add the supported_node_type variable

jt-edb:
With this commit, each role is able to define its own list of
supported node types.

### DIFF
--- a/roles/init_dbserver/tasks/main.yml
+++ b/roles/init_dbserver/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-# tasks file for edb.postgres.install role
+# tasks file for the init_dbserver role
 
 # reference epas variables
 - name: Reference pg_type variables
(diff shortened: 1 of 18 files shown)


vibhorkumar123:
@jt-edb This commit looks good. Do you think we should also variable the hardcoded pgbouncer part
```
item.value.node_type in supported_node_type) or item.value.node_type == 'pgbouncer'
```

jt-edb:
@vibhorkumar123 pgbouncer case is special because its roles can be applied only according to these conditions:
- `node_type=primary` **AND** `pgbouncer=true`
- `node_type=standby` **AND** `pgbouncer=true`
- `node_type=pgbouncer`

That's why checking only `supported_node_type` is not sufficient here.

vibhorkumar123:
Thank you, @jt-edb !

PR #76, opened 2020-12-15
Making setup_replication and setup_efm roles to use new format

vibhorkumar123:
Making setup_replication and setup_efm roles to use the new format


vibhorkumar123:
@jt-edb Can you please review the above changes?

roles/manage_dbserver/tasks/generate_password.yml
```diff
@@ -5,26 +5,34 @@
     mode: 0700
     state: directory
   become: no
+  run_once: true
+  no_log: "{{ disable_logging }}"
   delegate_to: localhost
 
 - name: Create passfile location
   set_fact:
-    passfile: "{{ pass_dir }}/{{ inventory_hostname }}_{{ input_user }}_pass"
```
jt-edb, line 12: When I introduced the node name into the pass file name, my goal was to ensure to have different passwords (for the same user name) when deploying multiple primary nodes or pem servers.

Do you think this is fine if we use the same random password for every nodes ?

roles/setup_efm/README.md
```diff
@@ -158,6 +166,7 @@ All the variables are available at:
 # EFM version 4.0
 $ ansible-playbook playbook.yml \
   -u centos \
+  -i inventory.yml \
   --private-key <key.pem> \
   --extra-vars="os=CentOS7 pg_version=13 pg_type=PG efm_version=4.0"
```
jt-edb, line 171: The `os` variable does not need to be set anymore.

roles/setup_efm/README.md
```diff
@@ -166,6 +175,7 @@ $ ansible-playbook playbook.yml \
 # 3.10
 $ ansible-playbook playbook.yml \
   -u ec2-user \
+  -i inventory.yml \
   --private-key <key.pem> \
   --extra-vars="os=RHEL8 pg_version=12 pg_type=EPAS efm_version=3.10"
```
jt-edb, line 180: The `os`  variable does not need to be set anymore.

roles/setup_efm/tasks/efm_cluster_set_params.yml
```diff
@@ -2,24 +2,21 @@
 - name: Set the efm parameters
   include_tasks: efm_set_params.yml
   when:
-    - item.value.node_type in ['primary','standby','witness']
+    - supported_node_type is superset(group_names)
```
jt-edb, line 5: What do you think about adding this in the `main.yml` file:
```yaml
- name: Ensure the current node type is supported
  meta: end_host
  when:
   - supported_node_type is not superset(group_names)
```

roles/setup_efm/tasks/setup_efm.yml
```diff
@@ -13,20 +13,91 @@
...
+  loop: "{{ lookup('pg_sr_cluster_nodes', wantlist=True) }}"
+  loop_control:
+    loop_var: node
+  no_log: "{{ disable_logging }}"
+
+- name: Ensure we have efm user in primary database
+  import_tasks: create_efm_user.yml
+  run_once: true
+  no_log: "{{ disable_logging }}"
+  delegate_to: "{{ primary_public_ip }}"
```
jt-edb, line 79: Could we just check if the current node is a primary node instead of executing this for each nodes ?

roles/setup_replication/tasks/setup_replication.yml
```diff
@@ -3,42 +3,95 @@
...
+  when:
+    - node.node_type == 'standby'
+    - node.replication_type == 'synchronous'
+  no_log: "{{ disable_logging }}"
+
+- name: Update primary for replication
+  import_tasks: primary_settings.yml
+  run_once: true
+  no_log: "{{ disable_logging }}"
+  delegate_to: "{{ primary_public_ip }}"
```
jt-edb, line 57: Do we need to execute this for each standby ?

roles/setup_replication/tasks/setup_replication.yml
```diff
@@ -3,42 +3,95 @@
...
 - name: Build standby service check
   block:
     - import_tasks: create_directories.yml
     - import_tasks: pg_basebackup.yml
     - import_tasks: configure_node.yml
   become: yes
-  delegate_to: "{{ item.public_ip }}"
 
-- name: Reset the the variables based on the user input
+- name: Update primary for replication
```
jt-edb, line 80: `Update primary for synchronous replication`

roles/setup_replication/tasks/setup_replication.yml
```diff
@@ -3,42 +3,95 @@
...
+  import_tasks: primary_synchronous_param.yml
+  when: synchronous_standbys|length > 0
+  run_once: true
+  delegate_to: "{{ primary_public_ip }}"
```
jt-edb, line 84: Same comment here, do we need to execute this for each standby ?

jt-edb (commented):
Just dropped some comments about minor improvements.

roles/manage_dbserver/tasks/generate_password.yml
vibhorkumar123: I did that intentionally because I noticed for each standby replication user was having a different password. With this, I had included `run_once: true` in manage_dbserver/tasks/generate_password.yml`

roles/setup_replication/tasks/setup_replication.yml
vibhorkumar123: This will be running only once for the cluster. I have included `run_once: true`

roles/manage_dbserver/tasks/generate_password.yml
vibhorkumar123: also for replication, we are using one standard user `repuser`

roles/setup_replication/tasks/setup_replication.yml
jt-edb: Alright, nice.

roles/manage_dbserver/tasks/generate_password.yml
jt-edb: I see, thank you. I'll apply some changes on #77

roles/setup_efm/tasks/setup_efm.yml
vibhorkumar123: Sure

PR #77, opened 2020-12-15
Split setup_pem into setup_pemserver and setup_pemagent

jt-edb:
Those new roles are able to be executed using the new inventory.yml file.


vibhorkumar123 (commented):
Do you think we should use some meaningful names for variables?
For example, ihn can be host_name, and hv can be host_var, something like that.

roles/setup_pemagent/tasks/pem_agent_config.yml
```diff
@@ -12,9 +12,9 @@
     PEM_SERVER_PASSWORD={{ pg_pem_admin_password }} \
     {{ pem_agent_bin_path }}/pemworker \
          --register-agent \
-         --display-name {{ item.key }} \
+         --display-name {{ inventory_hostname }} \
          --pem-user {{ pg_pem_admin_user }} \
-         --pem-server {{ pem_server_private_ip }} \
+         --pem-server {{ hv.pem_server_private_ip }} \
```
vibhorkumar123, line 17: Same as mentioned about the short name versus meaningful name.

vibhorkumar123 (changes requested):

roles/setup_pemagent/tasks/pem_agent_config.yml
jt-edb: Indeed. bc36959 fixes this

PR #99, opened 2020-12-29
Improve firewalld usage

jt-edb:
Changes:
- Open the ports using firewalld only if the service is enabled and started
- Open Postgres port
- Open EFM ports
- Open PEM server port

### DIFF
--- a/roles/setup_efm/tasks/setup_efm.yml
+++ b/roles/setup_efm/tasks/setup_efm.yml
@@ -68,7 +68,7 @@
 
 - name: Prepare efm node list
   set_fact:
-    efm_nodes_list: "{{ efm_nodes_list + node.private_ip + ':7800 ' }}"
+    efm_nodes_list: "{{ efm_nodes_list + node.private_ip + ':' + efm_port | string + ' ' }}"
   when:
     - not use_hostname
   loop: "{{ lookup('pg_sr_cluster_nodes', wantlist=True) }}"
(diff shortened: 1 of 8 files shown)


roles/setup_efm/tasks/setup_efm.yml
```diff
@@ -116,6 +116,20 @@
   include_tasks: efm_cluster_set_params.yml
   no_log: "{{ disable_logging }}"
 
+- name: Gather service facts
+  service_facts:
+
+- name: Open EFM TCP ports 7800-7810
+  firewalld:
+    port: "7800-7810/tcp"
```
vibhorkumar123, line 124: Shall we use a variable instead of hardcoded port? Let's add a variable efm_port in defaults/main.

roles/setup_pemserver/tasks/setup_pemserver.yml
```diff
@@ -65,6 +65,20 @@
 - name: Update pem probes and alerts
   import_tasks: pem_server_probe_alert.yml
 
+- name: Gather service facts
+  service_facts:
+
+- name: Open PEM TCP port 8443
```
vibhorkumar123, line 71: Same here. Let's keep the port in a variable

vibhorkumar123 (changes requested):

roles/setup_efm/tasks/setup_efm.yml
jt-edb: Please see this new commit 597f3f1

roles/setup_pemserver/tasks/setup_pemserver.yml
jt-edb: We will address this point later.

PR #128, opened 2021-01-21
  Add the ability to add efm_parameters with intial configuration of EFM

vibhorkumar123:
With this commit, a user would be allowed to use efm_parameters with the initial configuration of the EFM.


jt-edb (approved):
LGTM

PR #143, opened 2021-02-03
PgPoolII integration with EFM

jt-edb:


roles/setup_efm/vars/EPAS.yml
```diff
@@ -26,3 +26,8 @@ pg_port: 5444
 
 # postgres service
 pg_service: "edb-as-{{ pg_version }}"
+
+# pgpool2 binary dir
+pgpool2_bin_path: "/usr/edb/pgpool{{ pgpool2_version }}/bin"
+pgpool2_version: 4.1
```
vibhorkumar123, line 32: pgpool2_version is not in PG.yml. Let's move this variable to defaults/main.yml

vibhorkumar123 (changes requested):

roles/setup_efm/vars/EPAS.yml
jt-edb: Done

jt-edb:
Battle tested with EPAS 12 / CentOS 8:

- [x] switchover
- [x] failover
- [x] standby shutdown

vibhorkumar123 (approved):

PR #147, opened 2021-02-08
Making changes in pgpool integration with EFM

vibhorkum:
Making changes in pgpool integration with EFM. With this commit, we have the following:
1. A new function has been added to is_running in pcp_control.sh. This function checks if local Postgres and remote pgpool are running or not and accordingly can be used to take actions.
2. attach_node_id and promote_node_to_master function will perform actions after checking the node's current status at the remote pgpool side.
3. Instead of using the efm parameter 'script.remote.post.promotion', we will use 'script.custom.monitor' parameter for pgpool integration. This mechanism gives a better check. EFM will be using this script to check the node's status on remote pgpool before making any changes.


jt-edb (approved):
LGTM

vibhorkumar123:
Thank you, @jt-edb ! Merged

PR #149, opened 2021-02-09
Fixes for efm pgpool integration and pgpool2 permissions

vibhorkum:
Due to missing PGDATABASE environment variablein the pcp_attach_all.sh and pg_pcp_health.sh scipts, pcp commands are logging the following message in postgres log:
efm@efm app=[unknown] FATAL:  no pg_hba.conf entry for host 'xxx.xxx.xxxx.xxx', user 'efm', database 'efm', SSL on 

This commit fixes that. 

Pgpool2 sr_user has missing permission. Added pg_monitor role to pgpool2 user


jt-edb (approved):
LGTM

PR #158, opened 2021-02-28
Fixes pg_sr_cluster_nodes lookup filter

jt-edb:
Handle the case when the current node is not part of any SR cluster
found.


vibhorkumar123 (approved):

vibhorkumar123:
LGTM

PR #162, opened 2021-03-02
Making setup_efm to support Debian and also added force_initdb

vibhorkum:
With this PR following are the changes:
1. Support Debian in setup_efm
2. force_initdb changes for setup_replication
3. forec_initdb changes init_dbserver


roles/setup_efm/tasks/prepare_hba_value_list.yml
```diff
@@ -1,21 +1,46 @@
...
+      ]
   when: not pg_ssl
+  loop: "{{ pg_cluster_nodes }}"
+  loop_control:
+    loop_var: node
+  run_once: true
+  no_log: "{{ disable_logging }}"
 
-- name: Prepare the list
+- name: Prepare hba non ssl list
```
jt-edb, line 26: I guess this is the "hba ssl" list, isn't it?

jt-edb, line 26: I guess this the "HBA SSL" list, isn't it?

jt-edb (commented):
LGTM! Only one minor comment. Thank you

roles/setup_efm/tasks/prepare_hba_value_list.yml
vibhorkum: fixed

jt-edb (approved):
LGTM

PR #163, opened 2021-03-02
pg_sr_cluster_nodes improvement

jt-edb:
Handle the case when there is no standby nodes.


mw2q:
This patch is working for me.

mw2q (approved):

vibhorkumar123 (approved):

PR #165, opened 2021-03-02
CentOS8/RHEL8 disable inbuilt PG module

vibhorkum:
Ansible combining the comment and command in shell module due to which in built postgresql module is not getting disabled. This commit fixes that


mw2q:
This patch works for me.

vibhorkumar123:
Thank you, @mw2q !
Merging the changes!

PR #166, opened 2021-03-03
More FQCN usage

jt-edb:


galaxy.template.yml
```diff
@@ -23,3 +23,5 @@ issues: "https://github.com/EnterpriseDB/edb-ansible/issues"
 dependencies:
   "community.postgresql": "*"
```
vibhorkumar123, line 24: Do we want to be explicit about the version number for these collections?

vibhorkumar123 (changes requested):

galaxy.template.yml
jt-edb: Fixed

vibhorkumar123 (approved):

Issue #176, opened 2021-04-27
pgpool2 issue


dougortiz:
Issue when deploying in AWS EDB-RA-2, terraform 0.14.17 and ansible 2.10.6:

`fatal: [primary1]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'edb_devops.edb_postgres.pgpool2_nodes'. Error was a <class 'KeyError'>, original message: 'pgpool2'"}`

jt-edb:
Hello,

Thank you. This is now fixed in `master` and `v3_3`. A new maintenance release -- 3.3.1 -- has been published on GitHub and Ansible Galaxy.

Closing this issue.

PR #177, opened 2021-04-29
Fix pgpool2_nodes lookup filter

jt-edb:
Return an empty list even if there isn't any pgpool2 node defined
in the inventory.


jt-edb:
Fixes #176

PR #184, opened 2021-05-14
Support Azure Database

mw2q:
Generalize the RDS specific paths as DBaaS since it can be shared with
Azure Database.

Azure Database has some unique usage with respect to how the username
needs to be generated in a connect strict as opposed to generate usage
in SQL statements.  HammerDB as well as some of the scripts needs to be
generated differently because of this.  There is currently an open
ticket with HammerDB to handle this.  This change addresses Azure
Database directly for ease of development.


roles/setup_hammerdb/templates/loader.tcl.template
```diff
@@ -11,6 +11,12 @@ diset connection pg_port {{ pg_port }}
 diset tpcc pg_count_ware {{ tprocc_warehouses }}
 diset tpcc pg_num_vu {{ ansible_processor_vcpus }}
 diset tpcc pg_raiseerror true
-diset tpcc pg_superuser {{ pg_superuser }}
+{% if azure_db_hackery %}
```
jt-edb, line 14: This variable does not look to be defined in `setup_hammerdb/defaults/main.yml`

roles/setup_hammerdbserver/tasks/install_hammerdb.yml
```diff
@@ -7,3 +7,22 @@
...
+# We need this Azure Database hack until this is resolved:
+# https://github.com/TPC-Council/HammerDB/issues/163
+
+- name: Create HammerDB patch for Azure Database
+  template:
+    dest: hammerdb-{{ hammerdb_version }}-azure-db.patch
+    src: hammerdb-{{ hammerdb_version }}-azure-db.patch.template
+    mode: 0644
+  become: no
+  when: azure_db_hackery is defined
```
jt-edb, line 20: I'd suggest to check if `azure_db_hackery` is `True`

jt-edb (commented):
Thank you

roles/setup_hammerdb/templates/loader.tcl.template
mw2q: Right, it's set on the fly when a `deploy` is executed: https://github.com/mw2q/postgres-deployment/blob/azure-db/edbdeploy/project.py#L1261

roles/setup_hammerdbserver/tasks/install_hammerdb.yml
mw2q: Done,

jt-edb (approved):
LGTM

PR #185, opened 2021-06-24
Adding support for EFM 4.2

vibhorkum:
Adding support for EFM 4.2

### DIFF
--- a/roles/setup_efm/tasks/setup_efm.yml
+++ b/roles/setup_efm/tasks/setup_efm.yml
@@ -23,6 +23,19 @@
 - name: Reference variables
   include_vars: "{{ pg_type }}_{{ ansible_os_family }}.yml"
 
+- name: Find the WAL dir source location
+  stat:
+    path: "{{ pg_data }}/pg_wal"
+  register: waldir
+
+- name: Ensure parent WAL dir has correct permissions
+  file:
+    path: "{{ waldir.stat.lnk_source|dirname }}"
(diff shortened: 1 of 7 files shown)


roles/setup_efm/tasks/setup_efm.yml
```diff
@@ -23,6 +23,26 @@
...
+    path: "{{ pg_data }}/pg_wal"
+  register: waldir
+
+- name: Find the source of WAL dir if it's symlink
+  shell: >
+    dirname "{{ waldir.stat.lnk_source }}"
+  when:
+    - waldir is defined
+    - waldir.stat.islnk
+  register: wal_src_location
```
jt-edb, line 37: I think you could extract parent folder path with the help of the `dirname` jinja filter:
```yaml
- name: Find the source of WAL dir if it's symlink
  set_fact:
     wal_src_location: "{{ waldir.stat.lnk_source | dirname }}"
  when:
    - waldir is defined
    - waldir.stat.islnk
```

jt-edb (commented):
Looks good, one minor comment.

roles/setup_efm/tasks/setup_efm.yml
vibhorkumar123: Interesting. Thank you for sharing this.

vibhorkumar123: Fixed

jt-edb (approved):

PR #201, opened 2021-07-27
Setup rng-tools

jt-edb:


vibhorkumar123:
@jt-edb Do we need this in our edb-ansible repo?

vibhorkumar123:
Not needed in edb-ansible

PR #204, opened 2021-07-30
Allow the pemagent to be installed on non-dbservers

vibhorkum:
Allow the pemagent to be installed on non-dbservers


jt-edb (approved):
LGTM

PR #207, opened 2021-07-31
dbt2: add tasks to update system packages

mw2q:
There is a library update that exists at some stage in the RHEL life
cycle that breaks some packages if they are not updated.  Unfortunately
package dependencies don't catch this and the most reliable way to not
run into this is to make sure the whole system is up to date.

One specific instance is with cmake.


vibhorkumar123:
@mw2q Can you please resolve the conflicts and confirm if it's ready for further review?

roles/install_dbserver/tasks/install_dbserver.yml
```diff
@@ -31,5 +31,9 @@
     - force_install is defined
     - force_install
 
+- name: Update packages
+  package: name=* state=present
+  become: yes
+
```
jt-edb, line 37: Updating all packages will dramatically increase deployment time. Do we really need to do that?

mw2q:
We do until the base image is updated on Red Hat 8 based distros because
there are some package dependencies that don't get resolved if some of the
packages are too old.  I'm not an expert on repository management, so maybe
there is a better way?

mw2q:
This branch has been updated to use the dbt2 rpm packages, removing all of the system wide updates.

jt-edb (approved):

PR #214, opened 2021-08-18
Allow to install and setup replication without using replication slots

pgstef:
Hi,

It's not always convenient to use replication slots when you already have archive/restore_command configured and using a dedicated backup tool (barman, pgbackrest,...).

Adding a `use_replication_slots` (default true, so no behavior change to be expected) option seems a fair and quick way to achieve this, for at least `init_dbserver` and `setup_replication` roles.

Thanks in advance for your review.
Kind Regards,
Stefan

### DIFF
--- a/roles/init_dbserver/defaults/main.yml
+++ b/roles/init_dbserver/defaults/main.yml
@@ -11,6 +11,7 @@ efm_cluster_name: "efm"
 efm_service: "edb-{{ efm_cluster_name }}-{{ efm_version }}"
 
 disable_logging: yes
+use_replication_slots: true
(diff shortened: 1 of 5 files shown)


vibhorkumar123:
Thank @pgstef! LGTM.

vibhorkumar123 (approved):

pgstef:
Thanks @vibhorkumar123!
Any idea when the new release will be created?

vibhorkumar123:
We are planning a new release by tomorrow morning.

pgstef:
Awesome! Thanks again :-)

PR #218, opened 2021-08-19
Debian fix setup_replication and manage_dbserver

pgstef:
This fix some path issues for debian in `setup_replication` and `manage_dbserver`.
It also cleans up some multi-instance support left-overs.


vibhorkumar123 (approved):
LGTM

PR #222, opened 2021-08-20
Adding the fix for barman archive

vibhorkum:
Bumping the version to 3.5.2 and fixing the archive problem of barman


dougortiz (approved):
Reviewed and approved

PR #225, opened 2021-08-23
Fixing the unix socket issues for EPAS on RedHat

vibhorkum:
1. Fixed the default unix_socket issues in the v3_2 branch, which may be a problem with the RedHat server restart.
2. updated the version number


jt-edb (approved):
LGTM.

PR #245, opened 2021-09-24
Fix barman server setup when getenforce is absent

### LINKED ISSUE #234: Make SELinux tests pass on Debian
Just execute the ansible block for SELinux if we are not on Debian. Trivial, but maybe collection authors will want to install and configure SELinux on debian instead ? 

Purpose is to allow an install on Debian, to start with.

See patch here:
[0001-Make-SELinux-tests-pass-on-Debian.tar.gz](https://github.com/EnterpriseDB/edb-ansible/files/7185158/0001-Make-SELinux-tests-pass-on-Debian.tar.gz)

jt-edb:
Fixes #234

### DIFF
--- a/roles/setup_barmanserver/tasks/create_user.yml
+++ b/roles/setup_barmanserver/tasks/create_user.yml
@@ -16,10 +16,28 @@
     home: "{{ barman_home }}"
   become: yes
 
+- name: Check if getenforce is installed
+  ansible.builtin.shell: command -v getenforce > /dev/null 2>&1
+  register: getenforce_installed
+  become: true
+  ignore_errors: true
+
+- name: Set default value for selinux_enforcing
+  ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


jt-edb:
Hi @klando,

This PR should fix #234 making barman server configuration possible when `getenforce` is not installed, whatever the OS.

vibhorkumar123 (approved):

klando:
> Hi @klando,
> 
> This PR should fix #234 making barman server configuration possible when `getenforce` is not installed, whatever the OS.

Hi Julien

I did like that in my first patch but I was not satisfied by the `ignore_errors: true`. Thus the ansible block to trigger based on os family.

I have no clue if it makes sense to even try to setup SELinux on Debian like distros.... but maybe getting an opinion here can drive future development in this area.

Thoughts ?

PR #252, opened 2021-10-01
PostgreSQL 14 support - part. 1

jt-edb:
This commit introduces PostgreSQL 14 support.

The EDB-RA-2 tests passed successfully with this commit.

Pgpool-II packages for PostgreSQL 14 are missing at this point.

On Debian, while EPAS 14 is not available, the postgresql-14-sslutils
package can't be installed...


roles/init_dbserver/defaults/main.yml
```diff
@@ -34,7 +34,7 @@ pg_service: "postgresql-{{ pg_version }}"
 pg_user_home: "/var/lib/pgsql"
 pg_bin_path: "/usr/pgsql-{{ pg_version }}/bin"
 pg_initdb: "{{ pg_bin_path }}/postgresql-{{ pg_version }}-setup initdb"
-pg_initdb_options: "--encoding=UTF-8 --locale=en_US.UTF-8 --auth=peer --auth-local=scram-sha-256 --auth-host=scram-sha-256 --data-checksums"
+pg_initdb_options: "-k --encoding=UTF-8 --locale=en_US.UTF-8"
```
vibhorkumar123, line 37: Why did we remove other options? We added these options intentionally.

roles/init_dbserver/vars/PG_RedHat.yml
```diff
@@ -16,7 +16,7 @@ pg_systemd_global_unit_file: "/usr/lib/systemd/system/postgresql-{{ pg_version }
 pg_user_home: "/var/lib/pgsql"
 pg_bin_path: "/usr/pgsql-{{ pg_version }}/bin"
 pg_initdb: "{{ pg_bin_path }}/postgresql-{{ pg_version }}-setup initdb"
-pg_initdb_options: "-k -A scram-sha-256 --encoding=UTF-8 --locale=en_US.UTF-8 --auth-local=peer --pwfile /dev/zero"
```
vibhorkumar123, line 19: Same here.

PR #298, opened 2022-03-02
setup_pemserver

vincentp7212:
Testing framework setup_pemserver

### DIFF
--- a/tests/tests/test_setup_pemserver.py
+++ b/tests/tests/test_setup_pemserver.py
@@ -1,11 +1,7 @@
 import pytest
 
 from conftest import (
-    load_ansible_vars,
-    get_hosts,
-    get_os,
     get_pg_type,
-    get_primary,
     get_pemserver,
     get_pg_version,
     get_pg_unix_socket_dir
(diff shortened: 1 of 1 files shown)


tests/tests/test_setup_efm.py
```diff
@@ -61,4 +62,34 @@ def test_setup_efm_pg_read_all_settings():
...
+    host = get_primary()
+    packages = [
+        'openjdk-8-jdk',
+        'bsd-mailx',
+        'edb-efm44'
+    ]
+
+    for package in packages:
+        assert host.package(package).is_installed, \
+            "Package %s not installed" % packages
```
jt-edb, line 95: Including changes not related to the main subject of the PR is not a good practice. Please open a new PR if you want to apply changes on EFM tests.

tests/cases/setup_pemserver/inventory.yml.j2
```diff
@@ -0,0 +1,13 @@
+---
+all:
+  children:
+    primary:
+      hosts:
+        primary1:
+          ansible_host: {{ inventory_vars['primary1_ip'] }}
+          private_ip: {{ inventory_vars['primary1_ip'] }}
```
jt-edb, line 8: I don't think we need a `primary` node if we are going to test PEM server only. Could you confirm?

jt-edb (changes requested):
Please see my comments.

tests/cases/setup_pemserver/inventory.yml.j2
vincentp7212: Does not need primary node

PR #304, opened 2022-03-17
setup_repo testing framework update

vincentp7212:
yum repolist no longer lists "PostgreSQL common RPMs for RHEL/CentOS" and instead lists "PostgreSQL common RPMs for RHEL / Rocky 8"


jt-edb (approved):
LGTM

PR #307, opened 2022-03-25
setup_pgbouncer testing framework

vincentp7212:
Includes the following tests:

- PGBouncer Packages are installed
- PGBouncer service is running
- Able to connect to PGBouncer database
- Able to select pgbouncer_admin
- Checking if port matches vars.json
- Checking if config file matches


jt-edb (approved):
LGTM

PR #308, opened 2022-03-30
manage_pgbouncer testing framework

vincentp7212:
Includes the following tests:

- test_db is successfully created
- pid file is successfully created and in the correct path
- auth file is successfully created and in the correct path
- correct number of pgbouncer users is created


jt-edb (approved):
LGTM

PR #310, opened 2022-04-05
Setup barmanserver testing framework

vincentp7212:
Includes the following tests:

- Check to see if barman packages were successfully installed
- Check to see if barman config file was successfully created
- Check to see if barman is a sudo user


jt-edb (approved):
LGTM

PR #321, opened 2022-04-12
setup_barman testing framework

vincentp7212:
Centos7/Debian10 Error with WAL file


vincentp7212:
Includes the following tests:

- barman user is successfully created in primary
- barman backup server is configured correctly
- primary server status is active when using barman status function
- barman home_dir is configured correctly

tests/tests/test_setup_barman.py
```diff
@@ -0,0 +1,72 @@
...
+    with host.sudo(pg_user):
+        query = "Select * from pg_user where usename = '%s'" % barman_user
+        cmd = host.run('psql -At -p %s -h %s -c "%s" postgres' % (port, socket_dir, query))
+        result = cmd.stdout.strip()
+
+    assert len(result) > 0, \
+        "Barman user does not exist in primary database"
+
+
+
```
jt-edb, line 72: It would be interesting to check if the `barman backup` command works

jt-edb (commented):
Is it possible to rename the folder to `setup_barman`, because there is currently a trailing `_`?
Could you add this to the Github actions?

jt-edb (approved):
LGTM

PR #327, opened 2022-04-25
Setup hammerdbserver testing framework

vincentp7212:
Includes the following tests:

- HammerDB tar file is downloaded
- HammerDB is successfully installed in root path


jt-edb (approved):
LGTM.

PR #328, opened 2022-04-26
Enable/Disable postgres core dump

vibhorkum:
Adding option for enabling/disabling of postgres core dump using parameter +pg_enable_core_dump

### DIFF
--- a/roles/init_dbserver/defaults/main.yml
+++ b/roles/init_dbserver/defaults/main.yml
@@ -4,7 +4,6 @@ os: ""
 pg_type: PG
 pg_version: 14
 pg_ssl: true
-
 pg_instance_name: main
 
 # EFM service information
@@ -16,6 +15,7 @@ disable_logging: yes
 use_replication_slots: true
 use_hostname: true
 update_etc_file: true
+enable_core_dump: false
 
 pass_dir: "~/.edb"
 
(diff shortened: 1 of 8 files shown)


mw2q:
Generally speaking this looks sane, but keep in mind my Ansible is less than basic.  I have few comments that are more cosmetic.

1. I think the pg* prefix before *core_dump variables should be more generically named because this is enabling core dumps for any user.
2. Similarly putting in the configuration in pg specific named tasks might be a little misleading, maybe better put into a separate generic operating system role.

I'm also not super familiar with how systemd works, so I don't know how the limit settings propagate from the init script once postmaster is spawned.  If you modify the limits.conf file instead, that would be OS agnostic, otherwise you would have to repeat this change for every distro that has a different init system, or even changes within the same init system if override files are changed again.

vibhorkumar123:
> Generally speaking this looks sane, but keep in mind my Ansible is less than basic. I have some comments that are more cosmetic.
> 
> 1. I think the pg* prefix before *core_dump variables should be more generically named because this is enabling core dumps for any user, in other words this is not a postgres specific configuration.
> 2. Similarly, putting in the configuration in pg specific named tasks might be a little misleading. Maybe it is better put into a separate generically named operating system role.

Made the changes as per the comment.

> 
> I'm also not super familiar with how systemd works, so I don't know how the limit settings propagate from the init script once postmaster is spawned. If you modify the limits.conf file instead then that would be more distro agnostic, otherwise you would have to repeat this change for every distro that has a different init system, or even changes within the same init system if override files are changed again (e.g. location, format).

jt-edb (approved):
LGTM

PR #330, opened 2022-04-27
Setup_hammerdb testing framework

vincentp7212:
Includes the following tests:

- load-tproc-c is installed in the correct path
- run-tproc-c is installed in the correct path
- max_connections matches the default max_connections
- max_wal_size matches the default max_wal_size 

Cannot be included in Github Action Workflow because setup_hammerdb (tshirt_size = small) assumes it has at least 32 GB of free memory. The default for docker is 8 GB.


jt-edb (approved):
LGTM

PR #335, opened 2022-05-05
Setup_dbt2 testing framework

vincentp7212:
Includes the following tests:

- dbt2 packages are installed into primary
- dbt2 user is successfully created in database


jt-edb (approved):
LGTM

PR #336, opened 2022-05-23
Move arithmetic and concat. inside jinja template

jt-edb:
Ansible 2.13 and forward does not permit to perform arithmetic and
concatenation operations outside of the jinja template.

Reported by: Stefan Fercot


vibhorkumar123 (approved):
LGTM

PR #337, opened 2022-05-30
Only install sslutils packages when pg_ssl is enabled

pgstef:
This allows trying early releases of PG/EPAS when sslutils packages aren't available yet.


jt-edb (approved):
LGTM!

PR #338, opened 2022-06-03
Add playbooks for DBT-3

mw2q:
This new playbook provides tasks to install and setup the dbt-3 kit.

### DIFF
--- a/roles/setup_dbt3/templates/dbt3_profile
+++ b/roles/setup_dbt3/templates/dbt3_profile
@@ -1,6 +1,6 @@
 export DSS_PATH="/tmp/dss"
-export DSS_QUERY="/usr/local/src/dbt3/queries/pgsql"
-export DSS_CONFIG="/usr/local/src/dbt3/src/dbgen"
+export DSS_QUERY="/usr/local/share/dbt3/queries/pgsql"
+export DSS_CONFIG="{{ pg_user_home}}/TPC-H_Tools_v3.0.0/dbgen"
 
 export DBGEN="$DSS_CONFIG/dbgen"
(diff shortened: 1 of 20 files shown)


roles/setup_dbt3/templates/dbt3_profile
```diff
@@ -0,0 +1,17 @@
+export DSS_PATH="/tmp/dss"
+export DSS_QUERY="/usr/local/share/dbt3/queries/pgsql"
+export DSS_CONFIG="{{ pg_user_home}}/TPC-H_Tools_v3.0.0/dbgen"
+
+export DBGEN="$DSS_CONFIG/dbgen"
+export QGEN="$DSS_CONFIG/qgen"
+
+export DBT3_DIR=/pgdata/dbt3
```
jt-edb, line 8: Is the `$PGDATA` location hardcoded here?

jt-edb (commented):
Looks good. Please check the minor comment I've made.

roles/setup_dbt3/templates/dbt3_profile
mw2q: Yeah.  The way the scripts work (which may need improvement) is that if a specific flag is specified with run-workload, it will create the database in the location specified by PGDATA in the environment.  If that is not used, PGDATA isn't used and the scripts will just try to connect to the database as specified by the other environment variables.

jt-edb: Alright, in that case I think you should use the `pg_data` variable here:
```bash
export DBT3_DIR={{ pg_data }}/dbt3
```

jt-edb (approved):
LGTM

PR #348, opened 2022-06-27
Make sure "{{ pg_user_home }}/.ssh" exists

mw2q:
This is important for ease of DBT-2 testing on systems where postgres
isn't installed but we want to use the same OS user as what is used on
the database system.


mw2q:
@jt-edb I've updated this PR to incorporate more changes, but it is ready for review now.

jt-edb (approved):
LGTM

PR #349, opened 2022-07-01
Fix os checking in DBT-* playbooks

mw2q:
os wasn't being set in these playbooks.  Just use ansible_os_family.  We
can check the major version when we actually support more than one major
version in these playbooks.


jt-edb (approved):
LGTM

PR #351, opened 2022-07-14
dbt2: install flamegraph tool

mw2q:
Also fix up using explicit module names in this playbook.


mw2q:
@jt-edb Another small patch that is ready for review.

PR #355, opened 2022-08-16
Adding support for EDB Repos 2.0 in setup_repo role

vibhorkum:
This PR adds the support of EDB Repos 2.0.
Details of EDB Repos 2.0 is given below:
```
* Currently available for EDB Enterprise and EDB Standard
* Token-based authentication
* Single, simple URL for all download types
* Currently supporting Red Hat Enterprise Linux (RHEL), Oracle Linux (OL), CentOS, Rocky Linux, AlmaLinux, Debian, Ubuntu, and SUSE Linux Enterprise Server (SLES)
```

### DIFF
--- a/roles/setup_repo/tasks/PG_Debian_setuprepos.yml
+++ b/roles/setup_repo/tasks/PG_Debian_setuprepos.yml
@@ -3,7 +3,8 @@
   stat:
     path: "{{ edb_auth_conf }}"
   register: auth_conf
-  when: os != 'Debian9'
+  when: >
+    os != 'Debian9' and enable_edb_repo|bool and repo_token|length < 1
   become: true
 
 - name: Build EDB auth conf
@@ -17,6 +18,7 @@
   when:
     - os != 'Debian9'
     - not auth_conf.stat.exists or auth_conf.stat.size == 0
+    - enable_edb_repo|bool and repo_token|length < 1
   become: true
 
 - name: Install gpg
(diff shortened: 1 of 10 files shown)


roles/setup_repo/tasks/PG_Debian_setuprepos.yml
```diff
@@ -36,13 +38,16 @@
     url: "{{ edb_deb_keys }}"
     state: present
   become: true
+  when: >
+    enable_edb_repo|bool and repo_token|length <
```
jt-edb, line 42: should be: `enable_edb_repo and repo_token|length < 1`

roles/setup_repo/tasks/setup_repo.yml
```diff
@@ -27,8 +28,20 @@
...
-      msg: "repo_username = {{ repo_username }} or repo_password = {{ repo_password }} are not valid!."
-  when: enable_edb_repo|bool and (repo_username | length < 1 or repo_password | length < 1)
+      msg: "repo_username = {{ repo_username }} or repo_password = {{ repo_password }} or repo_token = {{ repo_token }} are not valid!."
+  when:
+    - enable_edb_repo|bool
+    - repo_token | length < 1 and (repo_username | length < 1 or repo_password | length < 1)
+
+- name: Define EDB repo script link for RedHat family
+  set_fact:
+      edb_repo_script_link: "https://downloads.enterprisedb.com/{{ repo_token }}/enterprise/setup.rpm.sh"
```
jt-edb, line 38: Wondering if this should be defined along with other variables in `vars/`

vibhorkumar123: Instead of `vars`, I used if condition for `edb_repo_script_link` in `defaults/main.yml`. 
The main reason for not using `vars` is that if we use it, we also need to focus on refactoring and reusing other variables based on `ansible_os_family`, which is a good amount of work, and we will pick up this work later and submit it as a different PR.

roles/setup_repo/tasks/PG_Debian_setuprepos.yml
vibhorkumar123: Fixed

jt-edb (approved):
LGTM

PR #359, opened 2022-08-23
pgpool2: disable builtin postgresql package

jt-edb:
On the os' from the redhat family in version 8, the default
postgresql meta package must be disabled if we to install packages
depending on the postgresql-X-server packages and coming from EDB
or PGDG repo.

Error was:

Problem: package pgpool-II-pcp-4.3.2-1.rhel8.x86_64 requires
postgresql14-server, but none of the providers can be installed

Reported by: Sergio Romera


vibhorkumar123 (approved):
LGTM!

PR #362, opened 2022-08-24
Use inventory_hostname with delegate_to

jt-edb:
Using ansible_host with delegate_to works only when using the SSH
connector and the default SSH port (22).

Using inventory_hostname is more portable and allows different SSH
port without needing additional SSH configuration.


vibhorkumar123 (approved):
LGTM

PR #368, opened 2022-08-29
Synchronization check in manage_dbpatches

vibhorkum:
Adding checks for synchronization of old primary before switchover. This PR also add capabilities where user can choose if they want to apply patches to only primary or standbys.


jt-edb (approved):
LGTM

PR #379, opened 2022-08-30
pgbackrest support

hannahms:
I added pgBackRest support to edb-ansible to provide more robust backup and recovery tool options. 

I added two roles, `setup_pgbackrest` and `setup_pgbackrestserver`.

#### `setup_pgbackrestserver`

This role configures a pgBackRest server node that stores the archives and backups of at least one PostgreSQL/EPAS node. The tasks include setting up the pgBackRest system user and group, creating directories needed, installing the pgBackRest package, generating SSH keys (to use when exchanging with each PostgreSQL/EPAS node) and generating and populating the pgBackRest configuration file. 

#### `setup_pgbackrest`

This role configures PostgreSQL/EPAS nodes (primary and standby(s)) to be able to backup and recover data to/from the pgBackRest server node. The tasks include installing the pgBackRest package, exchanging SSH keys with the pgBackRest server node to allow backups/recovery to take place via SSH connection, generating and populating the pgBackRest configuration file, configuring the respective PostgreSQL/EPAS settings on the primary and standby nodes and taking an initial backup. 

Users have the option to determine which method they want backups and archiving to use (standard archiving or asynchronous archiving), how many, if any at all, standby nodes are present (up to 3 allowed) and other variables available in the `main.yml` file of each respective role.


jt-edb (commented):
LGTM. Please check online comments.

PR #380, opened 2022-08-30
Remove setup_repo for BDR dependencies on BDR nodes. 

vibhorkum:
Removing the bdr nodes checks for setup_repo. Now the installation of BDR packages and setup_repo are based on install_bdr_packages and tpa_subscription_token


jt-edb (approved):
LGTM

PR #384, opened 2022-09-08
Allow usage of custom SSL key and certificate

jt-edb:
With the help of the new variables: pg_ssl_key_file, pg_ssl_cert_file, pg_ssl_ca_file and pg_ssl_crl_file, it is now possible to send and use existing SSL certificates/keys.


vibhorkumar123 (approved):
LGTM!

PR #385, opened 2022-09-08
Asynchronous standby fix for setup_replication

vibhorkum:
Fixing the failure occurs if you don't have only asynchronous standby properties set in the inventory file


jt-edb (approved):
LGTM

PR #388, opened 2022-09-12
Minor fixes of EFM eager failover

vibhorkum:
Added declaration of variables of efm_eager_failover in manage_dbpatches. Also, protected properties_changes from causing any failure due to no changes.


jt-edb (approved):
LGTM

PR #390, opened 2022-09-14
setup_efm: execute WAL checking op. with root priv

jt-edb:
Fixing a Permission denied error when the ssh_user does not have sufficient rights on PGDATA.


vibhorkumar123:
DBT2 tests are failing again. Can you please check and fix it?

vibhorkumar123 (approved):
LGTM!

PR #394, opened 2022-09-14
SSL and pg_init_conf_params

vibhorkum:
Adding the support for using the SSL local to managed nodes and allowing users to set certain parameters with init_dbserver


roles/init_dbserver/templates/postgresql.conf.template
```diff
@@ -33,3 +33,8 @@ autovacuum_vacuum_cost_limit = 3000
 idle_in_transaction_session_timeout = 10min
 shared_preload_libraries = '{{ pg_shared_libraries_list | join(",") }}'
 cluster_name = '{{ pg_instance_name }}'
+{% if pg_init_conf_params|length > 0 %}
+{% for item in pg_init_conf_params %}
+{{ item.name }} = {{ item.value }}
```
jt-edb, line 38: I'd recommend quoting the value here:
```yaml
{{ item.name }} = '{{ item.value }}'
```

roles/init_dbserver/defaults/main.yml
```diff
@@ -43,6 +43,8 @@ pg_initdb_options: "-k --encoding=UTF-8 --locale=en_US.UTF-8"
 pg_unix_socket_directories:
   - "/var/run/postgresql"
 
+# initial configuration for starting postgres
+pg_init_conf_params: []
```
jt-edb, line 47: Can you please add an example (commented) of how to define settings here?

jt-edb (commented):
LGTM. Please check my comments.

roles/init_dbserver/templates/postgresql.conf.template
vibhorkum: Wouldn't it treat everything, including boolean/number, as a string if we use quotes? This may lead to PostgreSQL startup failures.

roles/init_dbserver/defaults/main.yml
vibhorkum: Done

roles/init_dbserver/templates/postgresql.conf.template
vibhorkumar123: Fixed

jt-edb (approved):
LGTM

PR #396, opened 2022-09-15
efm-api-node-state support

jt-edb:
This commit adds support to the EFM api component on EL7.

For enabling it, the efm_api_node_state var must be set to true.


vibhorkumar123 (approved):

vibhorkumar123:
LGTM

PR #398, opened 2022-09-16
Remove standby limitation from setup_barman README.md

### LINKED ISSUE #397: Update setup_barman README.md
The README.md for the setup_barman role contains [the following statement](https://github.com/EnterpriseDB/edb-ansible/blob/e1feff7cdb0d7dbd9473529950caebc88dc7bc16/roles/setup_barman/README.md?plain=1#L75-L76):

> Note: `rsync` backups can only be proceeded on primary nodes. Backuping
> standby nodes must be done with the `postgres` backup method.

Although there were some limitations around backing up standby nodes with older versions of PostgreSQL there are no such limitations with currently supported PostgreSQL versions (PG10 and later).

This statement can therefore be removed.

mikewallace1979:
Removes the limitation around using `backup_method = rsync` against standby nodes. It is now possible to backup standby nodes using `backup_method = rsync` with all currently supported PostgreSQL versions so the limitation no longer applies. This is true as long as `backup_options = concurrent_backup` which is the case in edb-ansible for both rsync and postgres backup methods.

Closes #397.

### DIFF
--- a/roles/setup_barman/README.md
+++ b/roles/setup_barman/README.md
@@ -72,6 +72,8 @@ Backup method. Can be:
 
 Default: `rsync`
 
+Note: When backing up standby nodes with `backup_method = rsync` the minimum supported version of Barman is 2.8.
+
 ## Dependencies
 
 This role does not have any dependencies, but packages repositories should have
--- a/roles/setup_barman/README.md
+++ b/roles/setup_barman/README.md
@@ -72,9 +72,6 @@ Backup method. Can be:
 
 Default: `rsync`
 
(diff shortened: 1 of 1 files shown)


vibhorkumar123:
@mikewallace1979 Is there any version of barman which doesn't support this functionality? If yes, can you please highlight versions that support it and versions which don't.

Thank you!

mikewallace1979:
> @mikewallace1979 Is there any version of barman which doesn't support this functionality? If yes, can you please highlight versions that support it and versions which don't.
> 
> Thank you!

This functionality is supported since Barman `2.8` - I'll update the PR to include this minimum version requirement.

PR #409, opened 2022-10-03
add ubuntu support to pgbouncer roles

hannahms:
Passed tests with PG 12, 13, and 14 and EPAS 14. EPAS 12 and 13 are returning an error in the `init_dbserver` role so I skipped them for now.


hannahms:
Passed with EPAS 13, EPAS 12 not supported with Ubuntu20

jt-edb (approved):
LGTM

PR #411, opened 2022-10-04
fix epas < 14 install

hannahms:
Adds the commands to drop default database before the variable `install_package` is called again.


jt-edb (changes requested):
As seen together, this PR requires changes.

PR #413, opened 2022-10-31
Allow pemagent only configuration

vibhorkum:
This commit allows users to configure pemagent in case the user doesn't have access to ssh to the PEM server. To make it possible, we have added the following options:
1. node_ssh_to_pemserver: true
2. enable_pemagent_barman: true

### DIFF
--- a/roles/setup_pemagent/defaults/main.yml
+++ b/roles/setup_pemagent/defaults/main.yml
@@ -14,6 +14,8 @@ pg_pem_agent_password: ""
 pg_ssl: true
 disable_logging: true
 
+node_ssh_to_pemserver: true
+enable_pemagent_monitor_barman: false
 pem_server_exists: false
 # efm information
 efm_version: "4.5"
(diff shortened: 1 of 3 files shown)


roles/setup_pemagent/defaults/main.yml
```diff
@@ -14,6 +14,8 @@ pg_pem_agent_password: ""
 pg_ssl: true
 disable_logging: true
 
+node_ssh_to_pemserver: true
+enable_pemagent_barman: true
```
mw2q, line 18: I don't think this variable name makes sense.  PEM is EDB software and Barman is open source.  Perhaps it makes more sense to have a variable for each?

mw2q, line 18: I'm wondering if "enable_pemagent_monitor_barman" is a little clearer that we're enabling pemagent to monitor barman.  Or something like that. Maybe just "pemagent_monitor_barman"?

vibhorkumar123: Good suggestion. Changed to `enable_pemagent_monitor_barman`

mw2q (approved):
After making comments and Vibhor addressing them, I think this ready.

PR #415, opened 2022-11-02
dbt2 related udpates

mw2q:
Bump the dbt2 version to [v0.48.4](https://github.com/osdldbt/dbt2/blob/v0.48.4/ChangeLog), which contains fixes for capturing postgresql database statistics from DBaaS systems.

The DBT2 driver system also needs to have psql installed.


jt-edb (approved):
LGTM

PR #416, opened 2022-11-04
Pgpool ubuntu

hannahms:
Added ansible.builtin to all tasks, changed become: true for all tasks, and added ubuntu and debian support for all pgpool2 roles. With PG: checked for PG 13 and 14. With EPAS: checked for EPAS 14; EPAS 13 initialization in Debian is broken.


roles/manage_pgpool2/tasks/manage_pgpool2.yml
```diff
@@ -1,59 +1,57 @@
...
-  set_fact:
+  ansible.builtin.set_fact:
     os: "{{ ansible_distribution | replace('RedHat', 'RHEL') }}{{ ansible_distribution_major_version }}"
 
 - name: Reference pg_type variables
-  include_vars: "{{ pg_type }}.yml"
+  ansible.builtin.include_vars: "{{ pg_type }}_{{ ansible_os_family }}.yml"
+
+#- name: Reference pg_type variables
+#  ansible.builtin.include_vars: "../../setup_pgpool2/vars/{{ pg_type }}_{{ ansible_os_family }}.yml"
```
jt-edb, line 11: You can remove this part.

roles/setup_pgpool2/tasks/pgpool2_setup.yml
```diff
@@ -96,94 +96,166 @@
...
+
+# make sure items in pgpool2_bin_path are executable without directly referencing pgpool2_bin_path in command
+- name: Add pgpool executables to {{ pgpool2_user }} $PATH from {{ pgpool2_bin_path }}
+  ansible.builtin.shell:
+    cmd: >-
+      cp -r {{ pgpool2_bin_path }}/. /bin/
+  when:
+    - pg_type == 'EPAS'
+    - ansible_os_family == 'Debian'
+  become: true
```
jt-edb, line 230: What problem does this task try to solve? Making hard copies of the binaries is not recommended because it will lead to problems when updating the package if the new binaries are not copied into /bin.
Binaries' locations can be configured through the system unit file if needed.

hannahms: When executing commands like `pcp_node_count`, the pgpool2 bin `/usr/edb/pgpool4.3/bin` is not on the `enterprisedb` user's `$PATH` and the command cannot be run without explicitly calling the bin, for example, `/usr/edb/pgpool4.3/bin/pcp_node_count`.

jt-edb: I don't think not having `pcp*` command available in the current PATH is an issue. Please remove this task, making hard copies of binaries is not recommended. Thank you!

jt-edb (approved):
LGTM. Thank you!

PR #420, opened 2022-11-08
fix linting issues for setup_pgbackrest role

hannahms:
Fixes linting issues in setup_pgbackrest role. Tests have been successful for PG/EPAS 14/13 for rocky8, centos7, debian10 and ubuntu20.


jt-edb (approved):
LGTM

PR #423, opened 2022-11-08
Lint fixes for dbt roles

mw2q:


mw2q:
I accidentally tested the release instead of this test and see that I have some errors to fix in the roles, even though linting passed.

jt-edb (approved):
LGTM

PR #429, opened 2022-11-10
Fixing linting issues with setup|manage_efm and setup_repmgr

vibhorkum:
Fixed lint issues with the following roles

setup_efm
manage_efm
setup_repmgr With this,
I have fixed specific issues with test cases


mw2q (approved):
This looks sane to me.

PR #438, opened 2022-12-07
update workflow tests

hannahms:
Update workflow tests in `push.yml` to include the `ANSIBLE_CORE_VERSION` environment variable. If it is not set, the test will be unable to build the `ansible-tester` docker image and it will fail.


jt-edb (approved):
LGTM

PR #439, opened 2022-12-15
Hammerdb update

hannahms:
Changes version of hammerdb to v4.6.
Removes touchstone tasks from both hammerdb roles. 
Adds support for Debian10 and Ubuntu20 for both hammerdb roles.


hannahms:
New lint version again :/

jt-edb (approved):
LGTM

PR #440, opened 2022-12-18
Update pgbackrest

hannahms:
Updated pgbackrest roles to allow more than 3 standby nodes. The tests for these roles was updated to ensure standby indexing was consistent and accurate. 
An index for nodes was added to the `/lookup/pgbackrest_nodes.py` lookup function to have the indexing required in `pgbackrest.conf` stay consistent across all nodes. Using this index value, configuration lines are created and added to the end of the file. 
On the standby nodes, this index value is accessible for `pgbackrest.conf` templating.

### DIFF
--- a/roles/setup_pgbackrest/tasks/configure_pgbackrest.yml
+++ b/roles/setup_pgbackrest/tasks/configure_pgbackrest.yml
@@ -21,6 +21,15 @@
   when:
     - use_hostname|bool
 
+- name: Set _pgbackrest_host when using hostname
+  ansible.builtin.set_fact:
+    _standby_index_var: "{{ item.index_var }}"
+  loop: "{{ standby_node_info }}"
+  when:
+    - standby_present is defined
+    - "'standby' in group_names"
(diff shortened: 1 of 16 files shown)


roles/setup_pgbackrest/tasks/configure_pgbackrest.yml
```diff
@@ -21,6 +21,15 @@
   when:
     - use_hostname|bool
 
+- name: Set _pgbackrest_host when using hostname
```
jt-edb, line 24: `name` does not reflect what the task is doing (wrong var. name)

jt-edb (commented):
LGTM. Only one minor comment.

roles/setup_pgbackrest/tasks/configure_pgbackrest.yml
hannahms: I have update this task `name` and added the `_standby_index_var` variable to the `Reset local variables` task.

PR #445, opened 2022-12-28
Ansible core upgrade 2.13

hannahms:
The breaking point between `ansible-core` 2.12 and 2.13 involved jinja templating performing concatenation and arithmetic operations. 

In the original line, `"{{ _pgpool2_node_list }} + [ '{{ server.hostname }}' ]"`, the default set `_pgpool2_node_list` was square brackets, `[]`. When concatenating `+ ['{{ server.hostname }}']` in `ansible-core` 2.12, the expected result of `['192.168.92.3']` was achieved through one loop. In `ansible-core` 2.13, the output of this operation was a list of all characters individually, `['[', ']', ' ', '+', ' ', '[', ' ', '1', '9', '2', '.', ...`. To modify to achieve the desired result, all operations must occur within the same jinja template, 
`{{ _pgpool2_node_list | default([]) + [server.private_ip | string] }}`. 

This operation only occurred in the `setup_pgpool2` role. All roles were tested with an `ansible-core` 2.13 PG Rocky8 instance. The `setup_pgpool2` role was tested with `ansible-core` 2.12 as well. 

When testing all roles, I received a TypeError on the `setup_patroni` role when calling the `etcd_cluster_nodes` lookup function. I modified the lookup function to run successfully.

### DIFF
--- a/plugins/lookup/etcd_cluster_nodes.py
+++ b/plugins/lookup/etcd_cluster_nodes.py
@@ -59,8 +59,8 @@ class LookupModule(LookupBase):
             filter_etcd_cluster_name = terms[0]
 
         # etcd can be deployed on BDR, barman and proxy nodes
-        for group in ( 'primary', 'standby', 'pemserver', 'pgbouncer', 'pgpool2',
-            'barmanserver', 'witness', 'proxy', 'pgbackrestserver', ):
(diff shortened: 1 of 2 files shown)


vibhorkumar123 (commented):
LGTM!

PR #446, opened 2023-01-02
Module linesinfile

jt-edb:
This in-house Ansible module is intended to provide better performance than lineinfile when multiple lines must be managed.

Instead of using a loop, with linesinfile, we can pass an array of lines. This way, the module is executed only once, whatever the number of managed lines is.

Most of lineinfile's attributes are supported: line, regexp, insertafter, insertbefore, state, backrefs, firstmatch, create, owner, group, mode.

A few attributes are not supported: backup and search_string

### DIFF
--- a/plugins/modules/README.md
+++ b/plugins/modules/README.md
@@ -1,0 +1,40 @@
+# Modules
+
+## edb_devops.edb_postgres.linesinfile
+
+This in-house Ansible module is intended to provide better performance than
+the built-in `lineinfile` module when multiple lines must be managed.
+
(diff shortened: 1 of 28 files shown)


vibhorkumar123 (commented):
LGTM! However, can you please add a README of this module in Plugins modules?

hannahms (approved):
LGTM

jt-edb:
> LGTM! However, can you please add a README of this module in Plugins modules?

Done.

vibhorkumar123 (approved):

PR #449, opened 2023-01-05
Fix linesinfiles: do not write useless changes

jt-edb:
When the regexp matches but the content is the same, in this case we don't have to do anything.


vibhorkumar123 (approved):
LGTM

PR #450, opened 2023-01-09
Add handlers

hannahms:
This PR adds handlers to roles where it is required. To determine where handlers needed to be placed, I relied on ansible-lint warning messages. 

The roles that required handlers to be added were `install_dbserver`, `manage_dbserver` and `setup_efm`. The `/handlers/main.yml` file was added to each of these roles.

There were many parts of the efm roles that were discussed being changed, including binding the efm service and pg service on Debian/Ubuntu OS's. However, this change was not possible due to the fact that in `/roles/init_dbserver/tasks/pg_setup_systemd.yml` the global systemd unit file is only initialized on RHEL OS's. It was also discussed to have configuration changes only occur in `manage_efm`, however, this change was attempted and tested and was not successful without a major overhaul of both the `setup_efm` and `manage_efm` roles. These changes were reverted. The `setup_efm` and `manage_efm` changes were tested on rocky8 with `use_eager_failover: true`.


vibhorkumar123 (approved):
LGTM

jt-edb (approved):
LGTM

PR #453, opened 2023-01-16
Fix registerdb with EFM for pemagent version 9.0

vibhorkum:
This commit ensures the pemagent service restart after EFM and DB registration


hannahms (approved):

jt-edb (approved):
LGTM

PR #461, opened 2023-01-31
Ansible core 2.14 update

hannahms:
This updates the entire repository to be compatible with `ansible-core` version 2.14. Some changes were made to fix broken tests, particularly with the barman roles. The `pg_service` lookup function was updated to include Debian OS's instead of just RedHat.


vibhorkumar123 (approved):
LGTM

PR #464, opened 2023-02-05
Update Barman and PG Tuner versions

hannahms:
Updates barman and barman-cli package version to 3.2.0-1 for EPAS to reflect the package versions offered in EDB repo. The pg-tuner package name was also updated to include Debian OS's consistent with the EDB repos. Tasks where the pg-tuner is directly called had `and ansible_os_distribution == 'RedHat'` removed from the conditional statement. Tested with EPAS and PG on Rocky8 and Debian10.


jt-edb (approved):
LGTM. Could you backport those changes into the v3.16 branch?

PR #467, opened 2023-02-06
Allow EDB credentials update

jt-edb:
In case of using wrong credentials, if we want to replace them, we first need to remove the .repo file, and then, remove the corresponding package.

If the .repo file is not removed prior to package deletion, then, package deletion is not possible due to wrong creds.

The variable force_repo must be set true.


vibhorkumar123 (approved):
LGTM!

PR #472, opened 2023-02-06
Update DBT docker-compose files

hannahms:
Updates the `docker-compose.yml` files for `setup_dbt2`, `setup_dbt2_driver`, `setup_dbt2_client`, `setup_dbt3` and `setup_dbt7`. To ensure these roles are compatible with Centos7, the install files for the respective roles were updated for EL7 compatibility. The `setup_hammerdb` `docker-compose.yml` and Makefile were updated to reflect the tests only having a `hammerdb1` node and no longer a `primary1` node as well. 
The tests for roles `setup_dbt2`, `setup_dbt2_driver`, `setup_dbt2_client`, `setup_dbt3` and `setup_dbt7` use the RPM package feature and thus are failing on Centos7, however, the RPM package presence is the only failing test.

### DIFF
--- a/roles/setup_dbt2/tasks/dbt2_install_packages_client.yml
+++ b/roles/setup_dbt2/tasks/dbt2_install_packages_client.yml
@@ -1,5 +1,5 @@
 ---
-- name: Install packages for DBT-2 Client
+- name: Install packages for DBT-2 Client for EL8
   ansible.builtin.dnf:
     name:
       - 'https://github.com/osdldbt/dbt2-packaging/releases/download/v{{ dbt2_version }}/dbt2-client-{{ dbt2_version }}-1.el8.x86_64.rpm'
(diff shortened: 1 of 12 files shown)


mw2q:
There's some additional work needed for these specific tests to run on anything other than RHEL8 derivatives.  An immediate correction would be to not run tests on anything other than a RHEL8 derivative.  My recommendation is to remove the non-RHEL8 derivatives, then add back other distributions as they become fully supported by edb-ansible. (i.e. this doesn't mean the other distros can't be used at all, but additional packaging and playbook effort is need for edb-ansible to support them in the same way RHEL8 derivatives are being supported now.)

hannahms:
I will move ahead with removing EL7 distributions. I am going to delete this PR and submit a new one with those revisions.

PR #473, opened 2023-02-07
[setup_repo] update ca-certificates to latest

pgstef:
Hi,

On CentOS7 default docker images, the ca-certificates package isn't updated preventing to install the pgdg package.
With the ansible role updating it to its latest version, the problem is fixed.

Kind Regards,
Stefan


jt-edb (approved):
LGTM. Thank you, Stefan.

PR #474, opened 2023-02-07
Fix barman and pgpool2 append errors

hannahms:
In `setup_barman` and `setup_pgpool2`, a jinja2 appending error was missed when updating to ansible-core 2.13. This fixes the error to allow the lists to be appended correctly. I have also double checked to make sure this issue is not found elsewhere.


vibhorkumar123 (approved):
LGTM!

vibhorkumar123 (approved):

PR #477, opened 2023-02-08
Update DBT roles

hannahms:
Removes documented compatibility of EL7 distributions with DBT roles. As of now, only EL8 is compatible.


mw2q:
This looks sane to me.  (except the merge commit. :) )

mw2q (approved):

PR #478, opened 2023-02-09
addition hba appending error

mto79:
In addition to @hannahms  also the fix for the default situation in task:
- name: I**nitialize the hba_entries variable** in update_pg_hba.yml

Now resulting in :

`ok: [xxxxx] => {
    "msg": "[{'users': 'barman', 'databases': 'postgres', 'contype': 'hostssl', 'source': 'xxxxxx}, {'users': 'barman', 'databases': 'postgres', 'contype': 'host', 'source': 'xxxxxx'}] + [\n  {\n    'users': 'barman',\n    'databases': 'replication',\n    'contype': 'hostssl',\n    'address': '10.110.0.4/32'\n  },\n  {\n    'users': 'barman',\n    'databases': 'replication',\n    'contype': 'host',\n    'source': 'xxxxxx'\n  }\n]"
}`

Which is not correct!

Strange that the replication role has this more neat in a prepare_hba_value.yml ! Would advice to make this uninform.

### DIFF
--- a/roles/setup_barman/tasks/update_pg_hba.yml
+++ b/roles/setup_barman/tasks/update_pg_hba.yml
@@ -7,13 +7,13 @@
           'users': 'barman',
           'databases': 'postgres',
           'contype': 'hostssl',
-          'source': '{{ hostvars[inventory_hostname].barman_server_private_ip | default("127.0.0.1") }}/32'
+          'source': hostvars[inventory_hostname].barman_server_private_ip | default("127.0.0.1") + '/32' | string
         },
         {
           'users': 'barman',
           'databases': 'postgres',
           'contype': 'host',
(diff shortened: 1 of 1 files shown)


hannahms:
Hi @mto79,
Are you currently getting that result when you run the task?

mto79:
Hi @hannahms , i did get the error with all the two statements, and i saw your fix from 2 days ago. But when the when clause is there it just keeps the above only for the database postgres .. so yes that keeps than the result only for than not for the replication database hba rule. (using ansible 2.14 )

But you als made a change for the source:
So the rule 
'source': hostvars[inventory_hostname].barman_server_private_ip | default("127.0.0.1") + '/32' | string

Is also some thing you want to keep in line i think.

hannahms (approved):

hannahms:
I understand now, and I agree that keeping consistency is best. Thank you for your contribution!

PR #483, opened 2023-02-15
I want to install repmgr

z3473:


hannahms:
Hi @z3473,
We previously had support for repmgr included in this repository and have reverted the removal of the role `setup_repmgr`.

PR #497, opened 2023-02-24
Add enable_epel_repo and enable_pgdg_repo flags

hannahms:
In cases when users have configured access to local repositories and do not require configuration of the EPEL and/or PGDG repos, the flags `enable_pgdg_repo` and `enable_epel_repo` have been added to the `setup_repo` role. When set to `false` the flags enable skipping tasks not required for configuration of the respective repo(s).


vibhorkumar123 (approved):
LGTM

PR #505, opened 2023-03-17
Update force_initdb force_replication and force_efm

hannahms:
Updates and enhances the `force_initdb`, `force_replication` and `force_efm` removal flag tasks. Uninstalls packages, removes configuration files, and reverts configuration settings.

### DIFF
--- a/roles/setup_replication/tasks/rm_replication.yml
+++ b/roles/setup_replication/tasks/rm_replication.yml
@@ -53,8 +53,8 @@
     path: "{{ pg_data }}"
     state: absent
   when:
-    - force_rm_data is defined
-    - force_rm_data
+    - force_rm_pg_data is defined
+    - force_rm_pg_data
   become: true
 
 - name: Remove Postgres log directory if exists
@@ -71,8 +71,8 @@
   become: true
   when:
     - pg_wal|length > 0
-    - force_rm_wal is defined
-    - force_rm_wal
+    - force_rm_pg_wal is defined
+    - force_rm_pg_wal
 
 - name: Remove postgres default data directory if exists
   ansible.builtin.file:
--- a/roles/setup_replication/tasks/rm_replication.yml
+++ b/roles/setup_replication/tasks/rm_replication.yml
@@ -34,7 +34,7 @@
 
 - name: Remove configuration directory on Debian family
   ansible.builtin.file:
-    path: /etc/postgresql/{{ pg_version }}/{{ pg_instance_name }}/
+    path: "/etc/postgresql/{{ pg_version }}/{{ pg_instance_name }}/"
     state: absent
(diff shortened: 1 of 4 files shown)


roles/setup_replication/tasks/rm_replication.yml
```diff
@@ -50,9 +68,54 @@
   become: true
   when: pg_wal|length > 0
 
+- name: Remove postgres default data directory if exists
```
vibhorkum, line 71: I suggest adding another flag for the removal of the data directory and pg_wal directory, as they contain user data and it is recommended for users to use this flag when deleting these directories. What do you think?

roles/setup_replication/tasks/rm_replication.yml
```diff
@@ -15,9 +16,10 @@
   ansible.builtin.systemd:
     name: "{{ pg_service }}"
     state: stopped
+    enabled: false
   when:
-    - ansible_facts.services[pg_service] is defined
-    - ansible_facts.services[pg_service].state == 'running'
+    - ansible_facts.services[pg_service + '.service'] is defined
```
vibhorkum, line 21: Do you think we should append the suffix '.service' to the service name specified in the variable, and revise the documentation to explain to users how to utilize the service name variable?

vibhorkum (changes requested):
Could you take a look at the comments? Apart from that, everything seems fine.

roles/setup_replication/tasks/rm_replication.yml
hannahms: The only place this suffix is necessary is when calling the `ansible_facts.services`, while the `ansible.builtin.systemd` only needs the currently defined `pg_service`. We also call the `pg_service` variable to build other service names so adding the `.service` suffix may require us to remove it in other places.

roles/setup_replication/tasks/rm_replication.yml
hannahms: I will add a `force_wal` and `force_data` flag

hannahms: Should this be added to the tasks removing `pg_default_data` and `pg_local_wal_archive_dir`?

vibhorkumar123: The directory is not for WAL archives, but rather for the actual WAL files location in Postgres. The archive location can be configured using archive_command or archive_library, as described in https://www.postgresql.org/docs/current/continuous-archiving.html. As we are currently utilizing the pg_wal and pg_data variables for a user-defined location, let's create new variables with the same name, but prefix with either `rm or `force_rm`

vibhorkumar123 (approved):
LGTM!

PR #512, opened 2023-04-02
Add validation role

hannahms:
- Updates `setup_pgbackrest` and `setup_pgbackrestserver` tests to only have 1 standby node. 2 standby's were enabled while adding capabilities to pgbackrest to support more than 3 standby's, now it is exhaustive to run tests.
- Need to update `postgresql_query` module to `manage_dbserver/tasks/execute_sql_scripts`
- Tests for `setup_pgpool2`, `setup_pgbouncer` and `setup_pemagent` need updates to psql commands
- Need to confirm backup tools are adequately validated for presence of wal files in archive repo.
- Some further thought into how pgbackrest validation is done. Currently, the tasks done while validating `setup_pgbackrestserver` require the completion of the `setup_pgbackrest` role, which hasn't executed at the time of running the validation tasks. Further tidying of `setup_pgbackrest`, `setup_pgbackrestserver` and `setup_repmgr`
- `setup_pemserver` has one task that seems unsafe to run in a production environment.


hannahms:
Adds validation after each role through using the flag `use_validation` or by applying to an already deployed cluster with the flag `validate_only`. Currently, `use_validation` is, by current default, set to `true` and `validate_only` is set to false throughout each respective role. I am inclined to believe that the `use_validation` default as `true` could lead to more errors with respect to failing the validation tasks, though these issues will be worked through to add modularity with a multitude of deployment configurations. By only being able to test on each defined test case, this modularity has not been tested
I feel as though using the ansible module `community.postgresql.postgresql_query` allows for more flexibility with respect to this particular role. I don't feel as though the `manage_dbserver/tasks/execute_sql_scripts.yml` adds anything that is marginally better than calling the `community.postgresql.postgresql_query` module directly, and calling the module allows for the `port`, `login_user` and `become_user` to have more flexibility. 
Within the `setup_pgpool2` and `manage_pgpool2` roles, the `pgpool2_users` variable was being overriden by `include_` statements because of the way the test variables file `vars.json` is included as `-e` extra variables, which overrides all other defined variable definitons. This has been corrected by allocating the `pgpool2_users` variable to not be defined by user specification or input, but by a separate `pgpool2_service_users` variable that is fed into the `pgpool2_users` variable when needed. 
Within the `setup_repmgr` role, validation tasks were failing because of the replication slots not created or recognized during the registration of a node. While the physical replication slot `standby1` is created during the `setup_replication` role, however repmgr does not utilize this physical replication slot and instead lists `repmgr_slot_1`, `repmgr_slot_2`, and so on and so forth, to label the physical replication slots expected to be utilized. For the standby's already setup, these physical slot name's are updated to correspond to the `inventory_hostname` of the respective node within the `repmgr_register_node.yml`.

hannahms:
All places where `community.postgresql.postgresql_query` was used have been updated to utilize `manage_dbserver/tasks/execute_sql_scripts`. In the case of `manage_dbserver`, an additional variable, `_pg_query`, was built to pass through to `pg_query` in places where queries required looped variables.

vibhorkumar123 (approved):
LGTM!

PR #514, opened 2023-04-13
Add remove_only flag and update rm_setup_repo

hannahms:
Adds a `remove_only` flag to all removal tasks to make allow users to run the playbook to only remove a previous attempt at or successful installation and configuration. 
Updates `rm_setup_repo` tasks to be consistent with what is currently in the respective OS family's `PG_setup_repo` installation and configuration.


vibhorkumar123 (approved):
LGTM

PR #519, opened 2023-05-01
change default ansible-core version to 2.13

hannahms:
Change the default ansible-core version from 2.12 to 2.13 as 2.12 has reached EOL.


vibhorkumar123 (approved):
LGTM!

PR #522, opened 2023-05-09
update hammerdb version variable setup

hannahms:
Update HammerDB variables to support a list of versions and allow version to change without having to hardcode certain variables in the testing setup.

### DIFF
--- a/roles/setup_hammerdb/defaults/main.yml
+++ b/roles/setup_hammerdb/defaults/main.yml
@@ -38,11 +38,5 @@ supported_os:
 use_patroni: false
 
 supported_hammerdb_versions:
+  - "4.7"
   - "4.6"
-  - "4.5"
-  - "4.4"
-  - "4.3"
-  - "4.2"
-  - "4.1"
-  - "4.0"
-  - "3.3"
--- a/roles/setup_hammerdb/defaults/main.yml
+++ b/roles/setup_hammerdb/defaults/main.yml
@@ -4,7 +4,7 @@
 validate_only: false
 use_validation: true
 
-hammerdb_version: 4.6
+hammerdb_version: 4.7
 hammerdb_filename: "HammerDB-{{ hammerdb_version }}-\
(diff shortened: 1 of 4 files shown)


tests/cases/setup_hammerdb/vars.json
```diff
@@ -1,5 +1,5 @@
 {
   "use_hostname": false,
-  "hammerdb": "/home/hammerdb/HammerDB-4.6",
-  "hammerdb_version": "4.6"
+  "hammerdb": "/home/hammerdb/HammerDB-4.7",
```
mw2q, line 3: I don't think we should force an upgrade at this time because we're in the middle of testing with a specific version.

But I do still see room for improvement.  This path should be based on whatever the hammerdb_version is.

This would also allow us to test with 4.7 before making a chance in default version.

mw2q (changes requested):

tests/cases/setup_hammerdb/vars.json
hannahms: I can make that change.

I was updating because I rewatched the Benchmark KT meeting and Julien mentioned he wasn't using edb-ansible because it didn't have the latest version of HammerDB so thought this was needed.

roles/setup_hammerdb/defaults/main.yml
```diff
@@ -34,4 +34,15 @@ supported_os:
   - Debian10
   - OracleLinux8
   - Ubuntu20
+
 use_patroni: false
+
+supported_hammerdb_versions:
```
mw2q, line 40: Let's leave this for 4.6 and 4.7 for now.  Since I believe those are the only two we have used thus far.

mw2q (changes requested):

roles/setup_hammerdb/defaults/main.yml
hannahms: updated!

mw2q (approved):

PR #525, opened 2023-05-16
fix: convert `pg_version` to string in template

bryan-bar:
`install_dbserver` role error while testing https://github.com/EnterpriseDB/edb-benchmarks/pull/64:
```yaml
TASK [edb_devops.edb_postgres.install_dbserver : Set package list for EPAS RedHat] ***
fatal: [postgres1]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ package_list | default([]) + [\n  'edb-as' + pg_version + '-server' | string, 'edb-as' + pg_version + '-server-core' | string,\n  'edb-as' + pg_version + '-server-contrib' | string, 'edb-as' + pg_version + '-server-libs' | string,\n  'edb-as' + pg_version + '-server-client' | string, 'edb-as' + pg_version + '-server-sslutils' | string,\n  'edb-as' + pg_version + '-server-indexadvisor' | string, 'edb-as' + pg_version + '-server-sqlprofiler' | string,\n  'edb-as' + pg_version + '-server-sqlprotect' | string\n] }}): can only concatenate str (not \"int\") to str. can only concatenate str (not \"int\") to str"}
```

### DIFF
--- a/roles/install_dbserver/tasks/validate_install_dbserver.yml
+++ b/roles/install_dbserver/tasks/validate_install_dbserver.yml
@@ -4,7 +4,7 @@
     package_list: >-
       {{ package_list | default([]) + [
         'glibc-common', 'ca-certificates', 'postgresql' + pg_version | string,
-        'postgresql' + pg_version + '-server' | string, 'postgresql' + pg_version + '-contrib' | string,
+        'postgresql' + pg_version | string + '-server', 'postgresql' + pg_version | string + '-contrib',
         'sslutils_' + pg_version | string
       ] }}
   when:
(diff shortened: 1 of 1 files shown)


vibhorkumar123 (commented):
LGTM!
As I am contemplating on this, do you think it would be beneficial to incorporate the declaration of package names in defaults/main.yml and utilize variables in tasks/*.yml?

hannahms:
> LGTM! As I am contemplating on this, do you think it would be beneficial to incorporate the declaration of package names in defaults/main.yml and utilize variables in tasks/*.yml?

The package name list generation is for the validation tasks, there are many different variations of packages needed depending on `os_major_version`, which `pg_version`, whether `pg_ssl = true` or not, etc. so I do not think defining the package names in the `main.yml` file would be a good idea. There would have to be a different list for each possible setup, which with EPAS/PG and RHEL/Debian along with all the other flag dependencies, would become exponential.

PR #527, opened 2023-05-16
Update PEM and READMEs

hannahms:
- Updates `setup_pemagent` role to only have `pem_agent_fetch_adm_password.yml` run if there is no user-defined `pg_pem_admin_password`.
- Updates `setup_pemagent` role within `pem_agent_install.yml` to only have the `pg-backup-api` package installed under conditional `when: pem_server_version_int|int >= 840 and enable_pemagent_monitor_barman` instead of previous `when: pem_server_version_int|int >= 840 or enable_pemagent_monitor_barman`. The task name has also been updated to differentiate it from the `Install required packages for PEM agent` task in the same file to `Install required packages for PEM agent when enable_pemagent_monitor_barman`.
- Updates `setup_efm` role by adding the variable `efm_init_parameters` to allow users to input parameters to set at initialization. The documentation is also updated to reflect this variable replacing the `efm_parameters` variable, which is now primarily used within the `manage_efm` role. 
- All `README`s have been updated to reflect up-to-date version and OS support, as well as to include the `when:` conditional in playbooks.
- Default `playbook.yml` file has been updated to include the `setup_repmgr`, `setup_pgbackrest` and `setup_pgbackrestserver` roles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,5 +415,6 @@ Authors:
   * Julien Tachoires
   * Mark Wong
   * Vincent Phan
+  * Hannah Stoik
 
 Contact: **edb-devops@enterprisedb.com**
--- a/README.md
+++ b/README.md
@@ -416,5 +416,6 @@ Authors:
   * Mark Wong
   * Vincent Phan
   * Hannah Stoik
+  * Bryan Barajas
 
 Contact: **edb-devops@enterprisedb.com**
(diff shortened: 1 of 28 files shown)


hannahms:
Lets squash all of these commits also, it makes it easier to go through the changes when I'm looking back for a release

PR #528, opened 2023-05-17
update manage_dbserver validation tasks with new lookup function

hannahms:
To ensure that boolean values are validated correctly within `pg_postgres_conf_params`, the lookup function `pg_conf_params_boolean` has been added to take boolean parameter values and return them as `on` for `true, on, yes` and `off` for `false, off, no`. The README has been updated to add comments about inputing parameter names and values as strings to make sure python parses the value correctly.


bryan-bar:
Ref: https://github.com/EnterpriseDB/edb-benchmarks/pull/64

vibhorkumar123 (approved):
LGTM!

PR #529, opened 2023-05-19
Rename and update setup_touchstone to setup_touchstone_tools

mw2q:
The tools are the separation of the stat collection and profiling tools, which used to be combined with the original random data generation tools.  We were previously only using the stat collection tools.

Install the AppImage instead of installing from source.  Set version to v0.6.1.


hannahms (approved):
LGTM

PR #532, opened 2023-05-25
Now when pg_ssl=false don't check the package.

Sagebati:
Hi,

Small contribution, when running the playbook with `pg_ssl` we don't install postgres-sslutils. But the check is still done.
I removed added a conditional, so now it works as expected.


hannahms:
Hi there,
Thank you for pointing to where that change needs to be made. However, while this works for PG Debian instances, it does not for RHEL and EPAS instances. I will be re-releasing with a comprehensive fix shortly.

PR #536, opened 2023-05-26
fix pg_bin_path in setup_barman role

hannahms:
Fixes the `pg_bin_path` variable within the respective `/vars/*` files.


vibhorkumar123 (approved):
LGTM!

PR #541, opened 2023-06-13
Updated setup_repo for reflecting the PGD5 repository

vibhorkum:


roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,33 @@
...
   changed_when: reposub.rc == '0'
-  when: >
-    tpa_subscription_token|length > 0
-    and install_bdr_packages|bool
-    and pg_version|int >= 14
+  when:
+    - enable_edb_repo|bool
+    - tpa_subscription_token|length > 0
+    - install_bdr_packages|bool
+    - pg_version|int == 14
```
hannahms, line 158: Is this only for pg_version == 14 or does it apply to 12-14? Are we planning on only supporting PGD on v14 and above?

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,33 @@
...
+  when:
+    - enable_edb_repo|bool
+    - repo_token|length > 0
+    - install_bdr_packages|bool
+    - pg_version|int >= 15
+    - bdr_version|int >= 5
+  loop:
+    - "x86_64"
+    - "noarch"
+
```
hannahms, line 180: Does this same process not need to apply to Debian OS as well? There is a similar `apt_key` and `apt_repository` module to work with

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
hannahms: If `pg_version == 14` is accurate, then we can change the curl line to read `curl -sS "{{ edb_2q_base_repo_link }}/{{ item }}/14/rpm" | bash`

hannahms (changes requested):
Other than the comments I made, can we ensure these changes are reflected in the corresponding `PG_RedHat_rm_repos.yml` and `PG_Debian_rm_repos.yml`?

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
vibhorkumar123: I haven't checked Debian yet, but we will add Debian support later. Right now, our focus is on Rocky/RedHat support.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
vibhorkumar123: We are currently not interested in supporting other versions for BDR4. I may consider discontinuing support for BDR 4 in the future. Our focus should be on supporting PGD 5 and later versions.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,33 @@
...
+    enabled: true
+  when:
+    - enable_edb_repo|bool
+    - repo_token|length > 0
+    - install_bdr_packages|bool
+    - pg_version|int >= 15
+    - bdr_version|int >= 5
+  loop:
+    - "x86_64"
+    - "noarch"
```
hannahms, line 179: Can we hardcode the noarch repo link? That should cover all x86_64 instances and reduce complexity.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,33 @@
...
+    - bdr_version|int == 4
   loop: "{{ edb_2q_repositories }}"
 
+- name: Install PGD packages repo
+  ansible.builtin.yum_repository:
+    name: "enterprisedb-postgres_distributed-{{ item }}"
+    description: "enterprisedb-postgres_distributed-{{ item }}"
+    file: "enterprisedb-postgres_distributed"
+    baseurl: "{{ edb_repo_base_download }}/postgres_distributed/rpm/el/{{ ansible_distribution_major_version }}/{{ item }}"
+    gpgkey: "{{ edb_repo_base_download }}/postgres_distributed/gpg.key"
```
hannahms, line 168: I think these should be a variable defined in `main.yml`, similar to other repo url's and their corresponding gpg key's. Something like `edb_pgd_rpm_repo_8`, `edb_pgd_rpm_repo_7`,  `edb_pgd_apt_repo` and `edb_pgd_gpg_key`?

vibhorkumar123: I believe using the ansible_distribution_major_version variable is a good idea instead of making it specific to the OS version. I can see the advantages of having both the edb_pgd_rpm_repo and edb_repo_base_download_gpg_key.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
vibhorkumar123: I am attempting to minimize code by reusing it for both situations. If we use hardcoded values, it will result in the addition of more code.

vibhorkumar123 (commented):
Added more comments

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
vibhorkumar123: I made several changes and created a new array variable named pgd_rpm_repo. This variable encompasses the repository name, description, and repository link for both x86_64 and noarch. I used the same code using pgd_rpm_repo. By doing this, we guarantee that we avoid duplicating the code and declare it in defaults/main.yml.

roles/setup_repo/defaults/main.yml
```diff
@@ -30,7 +31,18 @@ edb_2q_base_repo_link: "https://techsupport.enterprisedb.com/api/repository/{{ t
...
+edb_repo_script_link: "{{ edb_repo_base_download }}/enterprise/setup.{{ 'rpm' if ansible_os_family == 'RedHat' else 'deb' }}.sh"
+
+# EDB postgres_distributed RPM repo links
+pgd_rpm_repo:
+  - name: "enterprisedb-postgres_distributed-x86_64"
+    description: "EDB Postgres Distributed Repo For x86_64"
+    repo_link: "{{ edb_repo_base_download }}/postgres_distributed/rpm/el/{{ ansible_distribution_major_version }}/x86_64"
+  - name: "enterprisedb-postgres_distributed-noarch"
+    description: "EDB Postgres Distributed Repo For noarch"
+    repo_link: "{{ edb_repo_base_download }}/postgres_distributed/rpm/el/{{ ansible_distribution_major_version }}/noarch"
```
hannahms, line 44: Doesn't the noarch repo work on x86_64?

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,31 @@
...
+    gpgkey: "{{ edb_pgd_rpm_gpg_key }}"
+    repo_gpgcheck: true
+    enabled: true
+  when:
+    - enable_edb_repo|bool
+    - repo_token|length > 0
+    - install_bdr_packages|bool
+    - pg_version|int >= 15
+    - bdr_version|int >= 5
+  loop: "{{ pgd_rpm_repo }}"
```
hannahms, line 177: If the `pgd_rpm_repo` includes both x86_64 and noarch, wouldn't it either install the same repo twice or, if on an aarch64 host, fail when attempting to download the x86_64 repo?

hannahms:
I don't think we need to include both the x86_64 and noarch repo links, just the noarch will be sufficient on all architectures.

vibhorkumar123 (dismissed):

roles/setup_repo/tasks/PG_RedHat_rm_repos.yml
```diff
@@ -33,6 +33,7 @@
     - "/etc/yum.repos.d/edb.repo"
     - "/etc/yum.repos.d/edb.repo.rpmsave"
     - "/etc/yum.repos.d/enterprisedb-enterprise.repo"
+    - "/etc/yum.repos.d/enterprisedb-postgres_distributed.repo"
```
hannahms, line 36: Is the name of the repo `enterprisedb-postgres_distributed-x86_64.repo` and `enterprisedb-postgres_distributed-noarch.repo`? Is there way to do something like `enterprisedb-postgres_distributed-*.repo`?

vibhorkumar123: The yum.repos.d file is named enterprisedb-postgres_distributed.repo, and the entries in the file pertain to the x86_64 and noarch repositories.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -149,12 +151,31 @@
...
+    - install_bdr_packages|bool
+    - pg_version|int == 14
+    - bdr_version|int == 4
   loop: "{{ edb_2q_repositories }}"
 
+- name: Install PGD RPM packages repo
+  ansible.builtin.yum_repository:
+    name: "{{ item.name }}"
+    description: "{{ item.description }}"
+    file: "enterprisedb-postgres_distributed"
```
hannahms, line 166: Can both repositories have the same name? Wouldn't this create two files with `enterprisedb-postgres_distributed.repo`? Maybe use the `{{ item.name }}` value?

hannahms (changes requested):
Update `yum_repository` `file` value and the corresponding repo names in the removal tasks.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
hannahms: If the `file` parameter is not included, the module defaults to `name`, so removal of this parameter would be sufficient.

roles/setup_repo/tasks/PG_RedHat_rm_repos.yml
hannahms: Awesome, thank you for confirming!

hannahms (approved):
LGTM

Issue #537, opened 2023-05-30
Error when trying to setup pgbackrest server


Sagebati:
Hi, 

I'm trying to setup a pgbackrest server. but I get 

```
TASK [edb_devops.edb_postgres.setup_pgbackrestserver : Build pgbackrest_conf_lines] *****************************************************************************************************************************************************************************************************************************************
fatal: [util3]: FAILED! => {"msg": "'pgbackrest_standby_configuration' is undefined. 'pgbackrest_standby_configuration' is undefined"}
```

here is my playbook 

```yaml
- hosts: pgbackrestserver
  name: Deploy pgbackrest servers
  become: true
  gather_facts: true
  collections:
    - edb_devops.edb_postgres
  pre_tasks:
    - name: Initialize the user defined variables
      ansible.builtin.set_fact:
        pg_version: 14
        pg_type: "PG"
        repo_cipher_password: "password"
        repo_retention_full_type: "count"
        repo_retention_full: 2
  roles:
    - role: setup_repo
      when: "'setup_repo' in lookup('edb_devops.edb_postgres.supported_roles', wantlist=True)"
    - role: setup_pgbackrestserver
      when: "'setup_pgbackrestserver' in lookup('edb_devops.edb_postgres.supported_roles', wantlist=True)"
```

```yaml
all:
    vars:
        ansible_python_interpreter: /usr/bin/python3
        enable_edb_repo: false
    children:
        dbservers:
            hosts:
                util2:
                    ansible_host: util2
                    ansible_hostname: dbserver1
                    ansible_user: ubuntu
                    ansible_port: 22
                    private_ip: xxxxxxxxx
                    hostname: util2
                    pgbackrest: true
                    pgbackrest_server_private_ip: xxxxxxxx
                util3:
        pgbackrestserver:
            hosts:
                util3:
                    ansible_host: util3
                    ansible_hostname: pgbackrest1
                    private_ip: xxxxxxxxx
                    ansible_user: ubuntu
                    ansible_port: 22
                    hostname: util3
        primary:
            hosts:
                util2:
        standby:
            hosts:
                util3:
```

any ideas ?

hannahms:
Hi there,

[...]

PR #542, opened 2023-06-14
fix pgbackrest validation and configuration

hannahms:
Fixes error incurred in issue #537 where the variable `pgbackrest_standby_configuration` is undefined when called to loop in building the `pgbackrest_conf_lines` variable. The condition of `standby_present` was added to not build these lines when no standby is present, as the lines are only added to the configuration file if a standby is present. 

In addition this fix, the `validate_setup_pgbackrest` tasks were not working as expected. The SSH setup previously configured did not connect the pgbackrestserver to the standby in the test case. This was because of a `delegate_tasks` that had been thrown into the `exchange_ssh_keys` tasks. This line was removed. 

More commenting was added to the `define_node_variables` tasks to make each task have a unique name.

### DIFF
--- a/roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
+++ b/roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
@@ -74,7 +74,6 @@
   delegate_to: "{{ _pgbackrest_server_inventory_hostname }}"
   become: true
   changed_when: false
-  # delegate_facts: true
 
 - name: Add pg_owner SSH public key into pgbackrest_server known hosts
   ansible.builtin.known_hosts:
--- a/roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
+++ b/roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
@@ -72,9 +72,9 @@
   ansible.builtin.command: ssh-keyscan {{ _pg_host }}
   register: _pgbackrest_ssh_keyscan_output
   delegate_to: "{{ _pgbackrest_server_inventory_hostname }}"
-  delegate_facts: true
   become: true
   changed_when: false
+  # delegate_facts: true
 
 - name: Add pg_owner SSH public key into pgbackrest_server known hosts
   ansible.builtin.known_hosts:
--- a/roles/setup_pgbackrest/tasks/validate_setup_pgbackrest.yml
+++ b/roles/setup_pgbackrest/tasks/validate_setup_pgbackrest.yml
@@ -14,10 +14,14 @@
   register: pgbackrest_config_check_res
   changed_when: pgbackrest_config_check_res.rc == '0'
 
+- name: Debug
+  ansible.builtin.debug:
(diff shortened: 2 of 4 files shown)


roles/setup_pgbackrest/tasks/validate_setup_pgbackrest.yml
```diff
@@ -14,10 +14,14 @@
   register: pgbackrest_config_check_res
   changed_when: pgbackrest_config_check_res.rc == '0'
 
+- name: Debug
+  ansible.builtin.debug:
+    var: "{{ pgbackrest_config_check_res.stdout_lines }}"
+
```
vibhorkumar123, line 20: Is there a particular reason for including the "stdout_line" in the debug? Do you think that including it could potentially expose sensitive information to the log? If we need to debug output, I believe a similar result is achievable with the "-vvv" option.

roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
```diff
@@ -72,7 +72,6 @@
   ansible.builtin.command: ssh-keyscan {{ _pg_host }}
   register: _pgbackrest_ssh_keyscan_output
   delegate_to: "{{ _pgbackrest_server_inventory_hostname }}"
-  delegate_facts: true
```
vibhorkumar123, line 75: Is there a specific reason for removing the delegate_facts? I hope its removal won't have a negative impact.

vibhorkumar123 (changes requested):

vibhorkumar123:
Please update the README to include information about the new variable, standby_present.

roles/setup_pgbackrest/tasks/validate_setup_pgbackrest.yml
hannahms: This was added to debug errors thrown, I will remove

roles/setup_pgbackrest/tasks/exchange_ssh_keys.yml
hannahms: It was causing the `ssh-keyscan` command to only collect and register information from one of the primary/standby nodes, not both. This meant that (in the test case), only the primary information was included in the `known_hosts` file on the pgbackrestserver node, which was causing authentication errors in the validation checks when the server node was trying to connect to the primary and standby. Removing solved this issues and the `known_hosts` file was updated properly.

vibhorkumar123: Please update the README to include information about the new variable, standby_present or any other changes

hannahms:
The `standby_present` variable has been included in the role since the PGBR configuration update to allow more than 3 standby nodes. 
It is not a user defined variable and is only defined in the `define_node_variables.yml` file if a standby exists in the cluster, otherwise it is undefined. It is meant to be used as a flag for certain tasks to run that are only necessary with a standby present in the cluster.

PR #543, opened 2023-06-17
add manage_touchstone_tools

hannahms:
Manage touchstone tools role to perform eight actions, each now contained within their own yaml file in the `manage_touchstone_tools` tasks. Each task has required variables to run, which are all specified in the README. The `manage_touchstone_tools.yml` is designed to be able to copy and paste the `import_role` block into a benchmark playbook. Examples of how to use these tasks is also documented in the README.

Within the `setup_touchstone_tools` role, I added the installation of the `fuse-sshfs` package, as that is a requirement to run the appimage in RHEL distros.

### DIFF
--- a/roles/manage_touchstone_tools/README.md
+++ b/roles/manage_touchstone_tools/README.md
@@ -104,25 +104,13 @@ Database server host or docker directory. Same as `-h` in `psql` commands.
 
 Database server port. Same as `-p` in `psql` commands.
 
-#### `pg_owner`
-
-The database connection username. Same as `-U` in `psql` commands.
  
 #### How to include `start_db_stats` in your playbook
 
 Below is an example of how to put `start_db_stats` in your playbook.
 
 ```yaml
-# if libpq environment vars are set on the machine, only ts_output_dir is required
-- name: Include start_db_stats tasks to start pgsql stats data collection
-  ansible.builtin.include_role:
-    name: manage_touchstone_tools
-    tasks_from: start_db_stats
-  vars:
-    ts_output_dir: "/var/log/touchstone/db_run_1"
-    sec_bw_sample: 60
-
-# if libpq environment vars are not set or you wish to explicitly define them
+# if vars pg_database, pg_port are already set by ansible, they are not required to be passed in
 - name: Include start_db_stats tasks to start pgsql stats data collection
   ansible.builtin.include_role:
     name: manage_touchstone_tools
@@ -133,7 +121,6 @@ Below is an example of how to put `start_db_stats` in your playbook.
     pg_database: "postgres"
     pg_server_hostname: "primary1"
     pg_port: 5432
-    pg_owner: "postgres"
 ```
 
 ### `stop_db_stats`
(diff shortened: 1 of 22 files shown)


roles/manage_touchstone_tools/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+Below is an example of how to put `stop_db_stats` in your playbook.
+
+```yaml
+- name: Include stop_db_stats tasks to stop pgsql stats data collection
+  ansible.builtin.include_role:
+    name: manage_touchstone_tools
+    tasks_from: stop_db_stats
+  vars:
+    ts_output_dir: "/var/log/touchstone/db_stats"
```
mw2q, line 163: The `ts_output_dir` should be the only var required for stopping.  The others should not be needed.  I wanted to double check if you found otherwise.  If so, it's a bug. :)

roles/manage_touchstone_tools/tasks/stop_db_stats.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Stop ts pgsql-stat
+  ansible.builtin.shell: |
+    set -o pipefail
+    ts pgsql-stat \
+      -o {{ ts_output_dir }} \
```
mw2q, line 6: Same comment here, about only needing the specify the data collection directory.

roles/manage_touchstone_tools/tasks/start_db_stats.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Start ts pgsql-stat
+  ansible.builtin.shell: |
+    set -o pipefail
+    ts pgsql-stat \
```
mw2q, line 5: I feel that that the flags for database name, port and user should default to the system environment, and should only be explicitly set if the ansible variables are set.  That would match typical usage of apps that use libpq (e.g. psql).

I realize these roles are set to use root (`become: true`)... so maybe it is necessary to do this.  So then I'm wondering if they should match the naming schedule used in manage_dbserver. i.e. pg_database, pg_owner, etc.  I can't think of a reason why we'd want a different set of variable names.

hannahms: Take a look at how I set `start_db_stats` to see if this is sufficient for the first point. I changed the variable names to match the naming conventions used elsewhere

mw2q: :thinking: That's handling the two cases where all variables set, or no variables set right?  After some thought, I think I prefer how you were doing it originally.

If we are using `become: true` if the `pg_owner` is now set, for example, the it'll default to `root` (assuming nothing else in the environment overrides it), which isn't ideal.

We do want it to be run by root, because pidstat (or /proc) observes user permissions such that a non-root user can only gather it's own stats (or something like that.)

mw2q (changes requested):
I think just the version bump and reverting the parameter handling is good to merge.

hannahms:
I think there could be a better way to omit the variables, but let me know how these changes look and we can always update to omit variables in the future :)

mw2q (changes requested):
Oops, still need `-U {{ pg_owner }}`.  Sorry if my typos made that confusing.

hannahms:
No worries, added back in

mw2q (approved):
Looks good to me.

PR #544, opened 2023-06-19
add any_errors_fatal: true

hannahms:
Added to all playbooks, playbook-examples, test case playbooks and README playbooks.


vibhorkumar123 (approved):
LGTM!

PR #545, opened 2023-06-19
update pgbackrest documentation

hannahms:
Update pgbackrest documentation to only include playbook examples where both the `setup_pgbackrest` and `setup_pgbackrestserver` roles are included.


vibhorkumar123 (approved):
LGTM!

PR #554, opened 2023-07-03
update powertools for RedHat8

hannahms:
No powertools repo available in RedHat8, equivalent would be to run `sudo subscription-manager repos --enable codeready-builder-for-rhel-8-x86_64-rpms` but with the UBI base-image used, there is no subscription to manage. The rest of the tasks work as the powertools/codeready builder repo does not prohibit the EPEL repo in RedHat8.


vibhorkumar123 (approved):
LGTM!

PR #562, opened 2023-07-13
add meta file for install_from_source

hannahms:
Adds the meta file for new role `install_from_source`.

### DIFF
--- a/roles/install_from_source/meta/main.yml
+++ b/roles/install_from_source/meta/main.yml
@@ -1,0 +1,44 @@
+---
+galaxy_info:
+  author: EDB
+  description: Download and installation of software from source
+  company: "EnterpriseDB"
+
+  # If the issue tracker for your role is not on github, uncomment the
+  # next line and provide a value
+  # issue_tracker_url: http://example.com/issue/tracker
+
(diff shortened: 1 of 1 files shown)


roles/install_from_source/meta/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  # If this a Container Enabled role, provide the minimum Ansible Container version.
+  # min_ansible_container_version:
+
+  #
+  # Provide a list of supported platforms, and for each platform a list of versions.
+  # If you don't wish to enumerate all versions for a particular platform, use 'all'.
+  # To view available platforms and versions (or releases), visit:
+  # https://galaxy.ansible.com/api/v1/platforms/
+  #
+  platforms:
```
mw2q, line 31: I think we can set this for all, then fix up as we find problems.  I don't believe anything is distro specific at this time.

mw2q (approved):

roles/install_from_source/meta/main.yml
hannahms: I'll update that when I update all repo documentation, this setting doesn't impact anything

PR #563, opened 2023-07-18
setup_dbt2: remove flame graph installation

mw2q:
These are provided by the dbt2 appimage.  No longer need to install them separately.


hannahms (approved):
LGTM

PR #565, opened 2023-07-20
PER-845 - PGD CAMO commit scopes

dougortiz:
Added support for: 
- oidentd service
- ident authentication
- pg_hba settings for 'enterprisedb' user
- commit scopes setup configuration
- extra nodes removal from commit scope


roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -142,6 +142,14 @@
     state: present
   become: true
 
+# Exclusive to pgd setup authentication
+# Package is used for ident authentication
+- name: Install oidentd package if not available
```
hannahms, line 147: Same as above `Install oidentd package when install_pgd`

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -142,6 +142,14 @@
...
+  ansible.builtin.package:
+    name: oidentd
+    state: present
+  become: true
```
hannahms, line 151: Add `when: install_pgd`

hannahms: This package on RHEL systems is also only in the EPEL repo, so the conditional should actually be
```
when:
  - install_pgd
  - enable_epel_repo
```

roles/setup_pgd/defaults/main.yml
```diff
@@ -113,3 +113,12 @@ supported_pg_version:
 
 supported_pgd_version:
   - 5
+
+# Other PG Users
+enterprisedb_user: "enterprisedb"
```
hannahms, line 118: Why is this listed? Doesn't the `/vars/EPAS_RedHat` cover the `pg_owner: enterprisedb` case?

roles/setup_pgd/defaults/main.yml
```diff
@@ -113,3 +113,12 @@ supported_pg_version:
...
+
+# PGD CAMO
+# PGD Cluster Node To remove for CAMO
+# PGD CAMO configuration requires a maximum of two nodes
+pgd_node_names_to_remove: ["edb_primary3"]
```
hannahms, line 123: I still don't love this idea of removing the PGD node, if a max of two nodes is required for CAMO, whats the point of a HA cluster? Upon further looking through the role and PGD documentation, I ran into a few concerning things that I will bring up on Monday.

roles/setup_pgd/tasks/setup_pgd.yml
```diff
@@ -106,3 +106,18 @@
   ansible.builtin.import_tasks: validate_setup_pgd.yml
   when:
     - validate_only|bool or use_validation|bool
+
+# Exclusive to pgd setup authentication
+- name: Start service oidentd, if not started
+  ansible.builtin.service:
```
hannahms, line 112: should this be `ansible.builtin.service` or `ansible.builtin.systemd`?

hannahms (changes requested):
There are some odd whitespace issues within the `validate_setup_pgd.yml` and also linting failures. I would like to spend more time going over this role and determining why its not behaving the way it should, so there are some minor changes now but I will probably have more on Monday.

roles/setup_pgd/tasks/validate_setup_pgd.yml
```diff
@@ -193,3 +193,9 @@
     init_data_query_res: null
     unix_stat: null
     pg_tde_query_result: null
+
+# Exclusive to pgd setup authentication
+- name: Start service oidentd, if not started
+  ansible.builtin.service:
+    name: oidentd
+    state: started
```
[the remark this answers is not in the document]
mw2q: Yeah, while this might help the immediate need, this probably needs to go somewhere more general.  We should continue discussing this specific issue in that other ticket.  (Sorry don't have reference offhand.)

roles/setup_pgd/defaults/main.yml
mw2q: I agree, pg_owner should be used where you found a need to create enterprisedb_user.

roles/setup_pgd/defaults/main.yml
mw2q: I discussed this briefly with Doug, but we maybe need another sanity check.  If CAMO simply does not work when there is more than 2 data nodes, then we need to update our test to use 2 nodes. 

Having said that, I don't know if the edb-ansible role needs any changes just to build 2 nodes.  Or whether that is a significant development effort.

mw2q (commented):
I concur with Hannah's review.

PR #567, opened 2023-07-26
PER-845 - PGD CAMO commit scopes

dougortiz:
Added support for:

- oidentd service
- ident authentication
- pg_hba settings for 'enterprisedb' user under pg_owner
- commit scopes setup configuration


roles/setup_pgd/tasks/validate_setup_pgd.yml
```diff
@@ -48,15 +48,15 @@
       - ansible_facts.services[pg_service + '.service']['state'] == 'running'
     fail_msg: "The service {{ pg_service }} is not running."
     success_msg: "The service {{ pg_service }} is running."
-
+
```
mw2q, line 51: extra whitespace :)

roles/setup_pgd/tasks/validate_setup_pgd.yml
```diff
@@ -48,15 +48,15 @@
...
+    
 - name: Check if service pg_service is enabled
   ansible.builtin.assert:
     that:
       - ansible_facts.services[pg_service + '.service']['status'] == 'enabled'
     fail_msg: "The service {{ pg_service }} is not enabled."
     success_msg: "The service {{ pg_service }} is enabled."
   when: ansible_os_family == 'RedHat'
-
+
```
mw2q, line 59: extra whitespace :)

roles/setup_pgd/tasks/validate_setup_pgd.yml
```diff
@@ -106,7 +106,7 @@
     success_msg: "Postgres data_directory is linked to {{ pg_data }}"
   when:
     - init_data_query_res.results[0].query_result[0].setting != pg_data
-
+
```
mw2q, line 109: extra whitespace :)

roles/setup_repo/tasks/PG_Debian_setuprepos.yml
```diff
@@ -29,6 +29,15 @@
     state: present
   become: true
 
+# Exclusive to pgd setup authentication
```
mw2q, line 32: This actually isn't exclusive to pgd.  It affects everything as long as we allow the default pg_hba.conf file to be created with ident auth.  So we want this for everyone.

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -142,6 +142,17 @@
     state: present
   become: true
 
+# Exclusive to pgd setup authentication
```
mw2q, line 145: Ditto. This actually isn't exclusive to pgd.  It affects everything as long as we allow the default pg_hba.conf file to be created with ident auth.  So we want this for everyone.

mw2q (changes requested):

roles/setup_repo/tasks/PG_Debian_setuprepos.yml
```diff
@@ -29,6 +29,14 @@
     state: present
   become: true
 
+# Package is used for ident authentication
+- name: Install oidentd package when install_pgd
+  ansible.builtin.package:
+    name: oidentd
+    state: present
+  become: true
+  when: install_pgd
```
hannahms, line 38: The install shouldn't happen in the `setup_repo` role

hannahms: And the conditional should be the same as for RedHat
```
  when: 
    - install_pgd  
    - enable_epel_repo
```

roles/setup_pgd/tasks/setup_pgd.yml
```diff
@@ -106,3 +106,18 @@
   ansible.builtin.import_tasks: validate_setup_pgd.yml
   when:
     - validate_only|bool or use_validation|bool
+
+# Exclusive to pgd setup authentication
+- name: Start service oidentd, if not started
+  ansible.builtin.service:
+    name: oidentd
+    state: started
```
hannahms, line 114: Did you ever try updating the `pg_hba` file with `ident` auth and the `pg_ident` file to enable the ident authentication?

roles/setup_pgd/tasks/pg_hba_config.yml
```diff
@@ -57,6 +57,64 @@
...
+  ansible.builtin.set_fact:
+    pg_allow_ip_addresses: >-
+      {{ pg_allow_ip_addresses | default([]) + [
+        {
+          "users": pg_owner,
+          "source": node.private_ip + "/32",
+          "databases": "postgres"
+          },
+          {
+          "contype": "host",
```
hannahms, line 100: this should either be absent or read `"contype": "hostssl"`

roles/setup_pgd/tasks/pg_hba_config.yml
```diff
@@ -57,6 +57,64 @@
...
+          "users": pg_owner,
+          "source": node.private_ip + "/32",
+          "databases": "postgres"
+          },
+          {
+           "contype": "host",
+           "users": pg_owner,
+           "source": node.private_ip + "/32",
+           "databases": "postgres"
```
hannahms, line 80: if we are using ident auth, should the method be `ident`?

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
```diff
@@ -0,0 +1,158 @@
...
+    apply:
+      throttle: 1
+  vars:
+    pg_query:
+      - query: "SELECT bdr.add_commit_scope(commit_scope_name := '{{ pgd_groupcommit_scope_name }}',
+                                          origin_node_group := '{{ pgd_two_node_group_name }}',
+                                          rule := 'ANY 2 ( {{pgd_two_node_group_name}} ) GROUP COMMIT'
+                                          )"
+        db: "{{ pgd_cluster_database }}"
+    ignore_query_execution_error: true
```
hannahms, line 155: Why would we want this to be true? If the query fails, shouldn't it read an error?

roles/setup_repo/tasks/PG_Debian_setuprepos.yml
dougortiz: Where should this piece of code be installed then?

roles/setup_pgd/defaults/main.yml
```diff
@@ -113,3 +113,10 @@ supported_pg_version:
 
 supported_pgd_version:
   - 5
+
+# PGD Commit Scopes
+configure_pgd_camo: true
+configure_pgd_groupcommit: false
+pgd_camo_scope_name: "pgd_camo_scope"
+pgd_groupcommit_scope_name: "pgd_groupcommit_scope"
+pgd_two_node_group_name: "pgd_two_nodes"
```
hannahms, line 122: this would only allow for one `pgd_two_node_group_name` to be input, should we allow for a list? Or make it a list of dictionaries with keys like `commit_scope_type`, `commit_scope_group`, `commit_scope_name` to allow for more than one CAMO or group commit setup?

roles/setup_pgd/tasks/pg_hba_config.yml
```diff
@@ -57,6 +57,64 @@
...
 
+- name: Prepare hba non ssl list for user - pg_owner
+  ansible.builtin.set_fact:
+    pg_allow_ip_addresses: >-
+      {{ pg_allow_ip_addresses | default([]) + [
+        {
+          "contype": "host",
+          "users": pg_owner,
+          "source": node.private_ip + "/32",
+          "databases": "postgres"
```
hannahms, line 68: Can we use a parameter for the `databases` value? This only allows for the database to be named `postgres`

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
mw2q: agreed.

roles/setup_pgd/tasks/pg_hba_config.yml
mw2q: Yeah, I believe this should be it's own parameter.  We don't necessarily want the default database (pg_database?) also be the initial replicated database. Having said that, I don't recall whether the initial replicated database is hard coded...

roles/setup_repo/tasks/PG_Debian_setuprepos.yml
mw2q: I feel like this should at least be near where ever we define the postgres packages to install.

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
```diff
@@ -0,0 +1,158 @@
...
+  ansible.builtin.include_role:
+    name: manage_dbserver
+    tasks_from: execute_sql_scripts
+    apply:
+      throttle: 1
+  vars:
+    pg_query:
+      - query: "SELECT bdr.add_commit_scope(commit_scope_name := '{{ pgd_camo_scope_name }}',
+                                          origin_node_group := '{{ pgd_two_node_group_name }}',
+                                          rule := 'ALL ( {{ pgd_two_node_group_name }} ) ON durable CAMO DEGRADE ON (timeout=500s) TO ASYNC'
```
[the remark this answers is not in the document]
dougortiz: Had the same thoughts with this SQL statement.

This is a larger discussion because adding the support for the other options will impact the deadline, not to mention we do not know if these other ON options are buggy or working at all.

roles/setup_pgd/tasks/pg_hba_config.yml
dougortiz: This piece of code was already there. Will update it.

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
dougortiz: Will update and re-test.

roles/setup_pgd/defaults/main.yml
dougortiz: So far architecturally, a bdr node group should only host two nodes.

If we add more support for other commit scopes, it will directly impact the deadlines for these benchmarks.

This is another subject for a larger discussion that can directly impact the deadline of these benchmarks.

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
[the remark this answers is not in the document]
hannahms: I think we should take the time and implement comprehensively and thoroughly, especially since this is just one revision of the `setup_pgd` role and more need to be done, I think we should avoid implementing something that we know needs more work done to make dynamic. If it means pushing a deadline back a day or two to check the implementation works, I think that is better than having to go back in a few months to implement multiple parameters and have to re-familiarize yourself with commit scopes/PGD/the `setup_pgd` role.

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
hannahms: Please have that change reflected on all queries in this file, lines 50, 66, 82, 101 and 119

roles/setup_pgd/tasks/setup_pgd.yml
dougortiz: It's working currently. Do not know if it'll break things by adding it to the 'pg_hba'

roles/setup_pgd/tasks/pg_hba_config.yml
mw2q: I think since Doug has proofed that it worked, this is more of a question of whether we want to rely on the default values we set in manage_dbserver.manage_hba_conf, vs explicitly setting this knowing it is the same as the default values.

So I think it's ok to leave as is unless we want to be more explicit everywehre.

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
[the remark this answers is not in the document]
mw2q: i see from the documentation (https://www.enterprisedb.com/docs/pgd/latest/durability/commit-scopes/) that `visible` is the default, when not specified.  i think it makes sense to have a variable set to `visible` by default, and always explicitly set the confirmation level, where we want to be able to set it to `durable` for any specific testing we're doing.  e.g. setting up CAMO

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
```diff
@@ -0,0 +1,158 @@
...
+  ansible.builtin.include_role:
+    name: manage_dbserver
+    tasks_from: execute_sql_scripts
+    apply:
+      throttle: 1
+  vars:
+    pg_query:
+      - query: "SELECT bdr.add_commit_scope(commit_scope_name := '{{ pgd_groupcommit_scope_name }}',
+                                          origin_node_group := '{{ pgd_two_node_group_name }}',
+                                          rule := 'ANY 2 ( {{pgd_two_node_group_name}} ) GROUP COMMIT'
```
hannahms, line 152: `{{ pgd_two_node_group_name }}`

roles/setup_pgd/tasks/pgd_commitscope_configuration.yml
```diff
@@ -0,0 +1,158 @@
...
+    apply:
+      throttle: 1
+  vars:
+    pg_query:
+      - query: "SELECT bdr.add_commit_scope(commit_scope_name := '{{ pgd_camo_scope_name }}',
+                                          origin_node_group := '{{ pgd_two_node_group_name }}',
+                                          rule := 'ALL ( {{ pgd_two_node_group_name }} ) ON durable CAMO DEGRADE ON (timeout=500s) TO ASYNC'
+                                         )"
+        db: "{{ pgd_cluster_database }}"
+    ignore_query_execution_error: true
```
hannahms, line 101: should these all have `run_once: true` set to not duplicate on all nodes?

roles/setup_repo/tasks/PG_RedHat_setuprepos.yml
```diff
@@ -142,6 +142,16 @@
...
+# Package is used for ident authentication
+- name: Install oidentd package when install_pgd
+  ansible.builtin.package:
+    name: oidentd
+    state: present
+  become: true
+  when:
+    - install_pgd
+    - enable_epel_repo
+
```
hannahms, line 154: this is not necessary and should be removed

roles/setup_pgd/tasks/pg_hba_config.yml
```diff
@@ -57,6 +56,52 @@
...
+          "databases": pg_database
+          }
+        ] }}
+  when: pg_ssl
+  loop: "{{ pgd_cluster_nodes }}"
+  loop_control:
+    loop_var: node
+  run_once: true
+  no_log: "{{ disable_logging }}"
+
```
hannahms, line 104: this is not necessary and should be removed

roles/init_dbserver/tasks/verify_pgd_settings.yml
```diff
@@ -0,0 +1,48 @@
+---
+- name: Check 'pgd_commit_scopes' dictionary length
+  ansible.builtin.assert:
+    that:
+      - pgd_commit_scopes | length > 0
+    msg: "The 'pgd_commit_scopes' list should have at least one dictionary."
```
hannahms, line 6: I don't understand why we would require having at least one commit scope. I think that these tasks `verify_pgd_settings.yml` should only be run if `pgd_commit_scopes | length > 0`

roles/setup_pgd/tasks/pgd_camo_commitscope_configuration.yml
```diff
@@ -0,0 +1,167 @@
...
+    ignore_query_execution_error: false
+  when:
+    - item.default_group_cs|bool
+    - item.cs_type|string == 'CAMO'
+    - _pgd_camoscope_exists|int == 1 and _pgd_camo_originnodegroup_exists|int == 1
+
+- name: ALTER max_prepared_transactions
+  community.postgresql.postgresql_query:
+    query: >-
+      ALTER SYSTEM SET max_prepared_transactions = 100
```
hannahms, line 153: Is there a way to allow users to set this parameter?

roles/setup_pgd/tasks/pgd_camo_commitscope_configuration.yml
```diff
@@ -0,0 +1,167 @@
...
+                                      )"
+        db: "{{ pgd_cluster_database }}"
+    ignore_query_execution_error: false
+  when:
+    - item.default_group_cs|bool
+    - item.cs_type|string == 'CAMO'
+    - _pgd_camoscope_exists|int == 1 and _pgd_camo_originnodegroup_exists|int == 1
+
+- name: ALTER max_prepared_transactions
+  community.postgresql.postgresql_query:
```
hannahms, line 151: Should this use the `manage_dbserver/execute_sql_scripts`?

roles/setup_pgd/tasks/rm_pgd.yml
```diff
@@ -10,57 +10,57 @@
...
+        pg_query:
+          - query: "SELECT bdr.part_node( node_name := '{{ pgd_local_node_name }}' )"
+            db: "{{ pgd_cluster_database }}"
+        ignore_query_execution_error: false
+      when:
+        - inventory_hostname != lead_primary_node.inventory_hostname
+        - _pgd_db_exists == 1
+        - _pgd_extension_exists == 1
+        - _pgd_local_node_created == 1
+        - _pgd_cluster_group_joined == 1
```
hannahms, line 54: this task needs to be apart of the block

roles/setup_pgd/tasks/setup_pgd.yml
```diff
@@ -106,3 +106,29 @@
...
+
+- name: Filter items based on cs_type 'GROUP_COMMIT'
+  set_fact:
+    filtered_items_group_commit: "{{ pgd_commit_scopes | selectattr('cs_type', 'equalto', 'GROUP_COMMIT') | list }}"
+
+- name: Display elements of each filtered 'GROUP_COMMIT' item
+  ansible.builtin.include_tasks: pgd_group_commitscope_configuration.yml
+  loop: "{{ filtered_items_group_commit }}"
+  loop_control:
+    label: "{{ item.cs_name }}"
```
hannahms, line 134: all of this needs to come prior to the `validate_setup_pgd` tasks and some tasks need to be added to `validate_setup_pgd` to validate the commit scope setup, if commit scopes were setup.

roles/setup_pgd/defaults/main.yml
```diff
@@ -113,3 +113,36 @@ supported_pg_version:
...
+#  for GROUP_COMMIT
+#  - cs_rule: "ALL ( pgd_remaining_nodes ) GROUP COMMIT"
+pgd_commit_scopes:
+  - cs_name: ''
+    cs_type: ''
+    parent_group: ''
+    cs_origin_node_group: ''
+    member_nodes: ['']
+    default_group_cs: true
+    cs_rule: ""
```
hannahms, line 148: I think these comments can be cleaned up a bit, I think one group commit example and one CAMO example is sufficient without breaking down each part. to leave `pgd_commit_scopes` empty, I think it needs to be set like `pgd_commit_scopes: ""`. All of this should take place prior to the last lines of the main.yml, where the `supported_pgd_version` var should be the last in the file.

roles/setup_pgd/README.md
hannahms, line 1: Can this be rebased to show up to date values? I have a hard time reading what has been changed vs. what is already changed in the main branch.

roles/init_dbserver/tasks/verify_pgd_settings.yml
```diff
@@ -0,0 +1,48 @@
...
+- name: Fail if more than one true value in default_group_cs
+  ansible.builtin.fail:
+    msg: "More than one true value found in default_group_cs"
+  when: true_count | int > 1
+
+- name: Check default_group_cs
+  assert:
+    that:
+      - pgd_commit_scopes | map(attribute='default_group_cs') | select("equalto", true) | list | count > 0
+    fail_msg: "At least one 'default_group_cs' value should be set to true."
```
hannahms, line 48: Why should this be the case?

roles/init_dbserver/tasks/verify_pgd_settings.yml
```diff
@@ -0,0 +1,48 @@
...
+  ansible.builtin.set_fact:
+    true_count: "{{ pgd_commit_scopes | map(attribute='default_group_cs') | select('equalto', true) | list | count }}"
+
+- name: Fail if more than one true value in default_group_cs
+  ansible.builtin.fail:
+    msg: "More than one true value found in default_group_cs"
+  when: true_count | int > 1
+
+- name: Check default_group_cs
+  assert:
```
hannahms, line 45: `ansible.builtin.assert`

hannahms (changes requested):

roles/init_dbserver/tasks/init_dbserver.yml
```diff
@@ -51,6 +51,13 @@
     - hostvars[inventory_hostname].pgd.use_physical_backup is defined
     - hostvars[inventory_hostname].pgd.use_physical_backup|bool
 
+- name: Import PGD verify tasks
+  ansible.builtin.import_tasks: verify_pgd_settings.yml
+  when:
+    - validate_only|bool or use_validation|bool
+    - not remove_only|bool
+    - install_pgd
```
hannahms, line 59: add `pgd_commit_scopes | length > 0`

PR #569, opened 2023-08-01
setup_dbt2: make sure hostssl rules exist for driver and client conne…

mw2q:
…ctions

This hopefully covers all default connection behavior by setting host and hostssl rules for any PostgreSQL clients programs used by DBT-2. The default behavior seems to vary for one reason or another.

Cannot use manage_dbserver.manage_hba_conf because it uses community.postgresql.postgresql_pg_hba, which does not let us create a host and hostssl line for the same source because contype is not part of the unique identifier that community.postgresql.postgresql_pg_hba uses.

### DIFF
--- a/roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
+++ b/roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
@@ -1,56 +1,59 @@
 ---
-- name: Gather DBT-2 client addresses
-  ansible.builtin.set_fact:
-    dbt2client_item: >-
-      {
-        'users': 'all',
-        'databases': 'all',
-        'contype': 'host',
-        'source': '{{ hostvars[item]['private_ip'] }}/32',
-        'method': 'trust'
-      }
+# Cannot use manage_dbserver.manage_hba_conf because it uses
+# community.postgresql.postgresql_pg_hba, which does not let us create a host
+# and hostssl line for the same source because contype is not part of the
+# unique identifier that community.postgresql.postgresql_pg_hba uses.
+
+- name: Find the hba file location
+  community.postgresql.postgresql_query:
+    login_user: "{{ pg_owner }}"
+    port: "{{ pg_port }}"
+    login_unix_socket: "{{ pg_unix_socket_directories[0] }}"
+    query: "show hba_file;"
+    db: "{{ pg_database }}"
+  become: true
+  become_user: "{{ pg_owner }}"
+  no_log: "{{ disable_logging }}"
+  register: hba_info
+
+- name: Update pg_hba.conf with DBT-2 client host connections
+  ansible.builtin.lineinfile:
+    path: "{{ hba_info.query_result[0].hba_file }}"
+    line: "host all all {{ hostvars[item]['private_ip'] }}/32 trust"
+    create: true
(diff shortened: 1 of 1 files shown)


roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
```diff
@@ -1,56 +1,55 @@
...
-  ansible.builtin.set_fact:
-    hba_entries: "{{ dbt2client_list.results | map(attribute='ansible_facts.dbt2client_item') | list }}"
+- name: Update pg_hba.conf with DBT-2 client host connections
+  ansible.builtin.lineinfile:
+    path: "{{ hba_info.query_result[0].hba_file }}"
+    line: "host all all {{ hostvars[item]['private_ip'] }}/32 trust"
+    create: true
+  with_items: "{{ groups['dbt2_client'] }}"
+  become: true
+  become_user: "{{ pg_owner }}"
```
hannahms, line 26: for benchmarking, does the use of `no_log: "{{ disable_logging }}"` matter at all? If so, I would recommend adding here, otherwise I'm okay with not having it

roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
```diff
@@ -1,56 +1,55 @@
...
-    pg_hba_ip_addresses: "{{ hba_entries }}"
+- name: Update pg_hba.conf with DBT-2 client hostssl connections
+  ansible.builtin.lineinfile:
+    path: "{{ hba_info.query_result[0].hba_file }}"
+    line: "hostssl all all {{ hostvars[item]['private_ip'] }}/32 trust"
+    create: true
+  with_items: "{{ groups['dbt2_client'] }}"
+  become: true
+  become_user: "{{ pg_owner }}"
   when: "'dbt2_client' in groups"
```
hannahms, line 37: just to confirm, the `dbt2_client` information you only want to be updated on `dbt2_client` nodes, but `dbt2_driver` information you want update on all `dbt2`/`dbt2_client`/`dbt2_driver` nodes?

roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
mw2q: for benchmarking, no.  for consistency with how we do things in the other roles, i will add momentarily.

roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
mw2q: yeah, iirc, the problem is that we don't have to have a client.  so if a client is not defined, we need to handle that accordingly.

we do always have to have a driver.  so if we don't have a driver defined then there are inherently a number of problems trying to use these playbooks. (having said that, i will admit i don't know if we handle that case of forgetting to define a driver well.)

roles/setup_dbt2/tasks/dbt2_update_pg_hba.yml
mw2q: forced pushed an update

dougortiz (approved):
LGTM

hannahms (approved):
LGTM

PR #570, opened 2023-08-02
setup_dbt2: add primary database ip to pg_hba.conf

mw2q:
There are cases where dbt2 scripts may access the primary database system by its private ip address locally.  Make sure it's allowed.


hannahms (approved):
LGTM

PR #571, opened 2023-08-03
minor updates to ensure patroni running smoothly

hannahms:
made some minor updates to ensure Patroni is running smoothly while benchmarking. 

- Updated `manage_dbserver/handlers/main.yml` to include a `become: true` parameter; this is required because the parameter `daemon_reload: true` requires root priviliges. 
- Updated `patroni.config.yml.j2` to set the postgresql `listen: "*"` as this parameter must match the `listen_addresses = '*'` parameter set in `postgresql.auto.conf` otherwise Patroni thinks you are trying to change the listen address.
- Updated `postgresql.conf.template` to sync up the `wal_level: logical` set in `patroni.config.yml.j2` and now is a default in the `init_dbserver` role. 
- Added `run_once: true` to more validate tasks and tasks within `patroni_generate_basic_password.yml` and `patroni_set_user_password_privs.yml`.
- Added an additional cluster restart within `primary_synchronous_params.yml` to ensure the cluster starts without the `restart required` label on primary.

### DIFF
--- a/roles/manage_dbserver/handlers/main.yml
+++ b/roles/manage_dbserver/handlers/main.yml
@@ -1,6 +1,8 @@
 ---
+# needs to be run with become: true because of daemon_reload
 - name: Reload the pg service
   ansible.builtin.systemd:
     name: "{{ pg_service }}"
     daemon_reload: true
     state: reloaded
+  become: true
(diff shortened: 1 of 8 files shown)


mw2q (approved):
Looks sane to me.

vibhorkumar123 (approved):
LGTM!

PR #574, opened 2023-08-04
add dynamic postgres params in patroni config

hannahms:
Certain postgres parameters must be set in the Patroni configuration file otherwise they will be overridden by Patroni default values. To allow customization of these parameters, the `patroni_pg_init_params` variable is added and the parameters listed, if any, are included in the PostgreSQL parameters section of the Patroni config file. 
The README has been updated to include information about the default settings, how and why to customize, what the resulting setting will look like in the config file, and links back to the Patroni documentation _Patroni configuation important rules_.

Under the `basebackup` method section in the Patroni config file, the `--create-slot` and `--slot inventory_hostname` `pg_basebackup` parameters have also been added because the log file showed it was recreating the slots perpetually, so explicitly listing out to create them removes the amount of junk in the log file.

### DIFF
--- a/tests/cases/setup_patroni/vars.json
+++ b/tests/cases/setup_patroni/vars.json
@@ -2,4 +2,14 @@
   "use_hostname": false,
   "pg_data": "/opt/pg_data/pgdata",
   "pg_wal": "/opt/pg_wal/pgwal",
+  "patroni_pg_init_params": [
+    {
+      "name": "wal_keep_segments",
+      "value": "32"
+    },
+    {
+      "name": "wal_keep_size",
+      "value": "160"
+    }
+  ]
 }
--- a/roles/setup_patroni/README.md
+++ b/roles/setup_patroni/README.md
@@ -27,20 +27,47 @@ The requirements for this ansible galaxy role are:
 
 When executing the role via ansible there are three required variables:
 
-  * ***os***
+### os
 
-  Operating Systems supported are: CentOS7, RHEL7, CentOS8, RHEL8, AlmaLinux8, Debian10 and Ubuntu20
+Operating Systems supported are: CentOS7, RHEL7, CentOS8, RHEL8, AlmaLinux8, Debian10 and Ubuntu20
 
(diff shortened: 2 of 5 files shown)


tests/cases/setup_patroni/vars.json
vibhorkumar123, line 1: This gets created dynamically. I think you should exclude this file

vibhorkumar123 (changes requested):

tests/cases/setup_patroni/vars.json
hannahms: This file passes in variables to the test case run through the `-e` extra parameters in the `ansible-playbook` command

hannahms: In the `/tests/docker/exec-tests.sh` script the playbook command is:
```
ANSIBLE_PIPELINING=1 ansible-playbook \
	-i /workspace/tests/cases/${CASE_NAME}/inventory.yml \
	--extra-vars "repo_username=${EDB_REPO_USERNAME}" \
	--extra-vars "repo_password=${EDB_REPO_PASSWORD}" \
	--extra-vars "repo_token=${EDB_REPO_TOKEN}" \
	--extra-vars "enable_edb_repo=${EDB_ENABLE_REPO}" \
	--extra-vars "pg_type=${EDB_PG_TYPE}" \
	--extra-vars "pg_version=${EDB_PG_VERSION}" \
	--extra-vars "@/workspace/tests/cases/${CASE_NAME}/vars.json" \
	--extra-vars "ansible_core_version=${ANSIBLE_CORE_VERSION}" \
	--private-key /root/.ssh/id_rsa \
	/workspace/tests/cases/${CASE_NAME}/playbook.yml
```
Without the `vars.json` file this command would no execute.

vibhorkumar123: Ah! I was thinking about the Patroni.json

vibhorkumar123 (approved):
LGTM!

PR #575, opened 2023-08-07
update dbt2 pgpass

hannahms:
Updates `setup_dbt2/dbt2_configure_pgpass.yml` to use `manage_dbserver/manage_pgpass` tasks to update the `~/.pgpass` file when it already exists. This keeps the DBT-2 role from overriding already configured values within the DB server pgpass. 
Keeps `azure_db_hackery` parameters.


vibhorkumar123 (approved):
LGTM!

PR #578, opened 2023-08-17
setup haproxy

hannahms:
- Updates etcd configuration with new yaml configuration file to allow for access to `etcdctl` without flags.
- Adds HAProxy setup for standalone server or alongside a patroni server.


vibhorkumar123 (approved):
LGTM!

bonesmoses:
Was there some particular reason to hard-code `maxconn 100` into each haproxy config line?

hannahms:
I set the global `maxconn` parameter to default to 100, but it can be changed by setting the `haproxy_global_maxconn` variable

bonesmoses:
I don't see that variable used in the "Build haproxy_server_lines" task. It's just injected into the HAProxy config itself. It's being overridden in each individual server line to be hardcoded to "maxconn 100". I guess I could be missing something, but that's how I read that task.

PR #579, opened 2023-08-21
setup_dbt2 dbt2_configure_pgpass: fix logic

mw2q:
Fix two issues:

Don't overwrite an exist .pgpass file.  This playbook previously wasn't taking into consideration that the .pgpass file it is editing was already exist of not, always override it with a template.  Using ansible.builtin.lineinfile is hopefully the best way to handle that.

Don't assume we're not a DBaaS system.  manage_dbserver playbooks will fail if we're setting up a system that is testing a DBaaS database.  The ansible.builtin.lineinfile task should be able to handle both cases appropriates.


hannahms (approved):
LGTM

PR #580, opened 2023-08-21
fix readme and update maxconn

hannahms:


roles/setup_haproxy/tasks/haproxy_configure.yml
```diff
@@ -9,7 +9,7 @@
     haproxy_server_lines: >-
       {{ haproxy_server_lines | default([]) + [
         "server " + node.inventory_hostname + " " + node.private_ip | string + ":" + pg_port | string + " " +
-        "maxconn 100 check port " + patroni_rest_api_port | string
+        "maxconn " + haproxy_global_maxconn | string + " check port " + patroni_rest_api_port | string
```
vibhorkumar123, line 12: I would recommend including a default value for `haproxy_global_maxconn`.

vibhorkumar123 (changes requested):

roles/setup_haproxy/tasks/haproxy_configure.yml
hannahms: default is set to 100 in the main.yml file

vibhorkumar123: Ok. LGTM!

vibhorkumar123 (approved):

PR #582, opened 2023-08-22
Add fio

mw2q:
* fix an issue with install_from_source
* add setup_fio playbook


hannahms (approved):
LGTM! Nice validation tasks :)

mw2q:
i added bc to the list of packages to install.  it's required depending on whether or not math is used in a configuration.

PR #586, opened 2023-08-31
Add rhel9/rocky9 support

hannahms:
Adds support for RHEL9 and Rocky9, when applicable. Some packages are not available in EL9 repos yet and `main.yml` `supported_os` reflects accurately.


vibhorkumar123 (approved):
LGTM!

PR #589, opened 2023-09-05
setup_pgd pgd_commit_scopes default empty list

mw2q:
The default value for pgd_commit_scopes should be an empty list as opposed to an empty string because all of the validations are written for lists as opposed to strings.


dougortiz (approved):
Looks good

vibhorkumar123 (approved):
LGTM!

PR #600, opened 2023-10-10
Move core dump setup into manage_operating_system

mw2q:
* Move handling of installing helper packages to install debug symbols into manage_operating_system role; running the tools to install the debug symbols stay near respective package installation tasks
* Set core size unlimited for all users, not just the database owner
* Set core size limit in limits.conf, not in systemd
* Do not adjust OS default setuid for potential core dumps
* Stop overriding core dump locations, which confuses distro specific helper tools

Note that core dump handling is currently only handled for RHEL based systems and EPAS.  These changes rearrange the functionality that is currently available.


hannahms:
Where do the operations for the following items fall?
> * Set core size unlimited for all users, not just the database owner
> * Set core size limit in limits.conf, not in systemd
> * Do not adjust OS default setuid for potential core dumps
> * Stop overriding core dump locations, which confuses distro specific helper tools

I know for the last item, not overriding the location essentially means not defining our own core dump directory, as we have been doing. However, I don't see any updates to the tasks within the `manage_operating_system` role?

hannahms (changes requested):

mw2q:
> Where do the operations for the following items fall?
> 
> > * Set core size unlimited for all users, not just the database owner
> > * Set core size limit in limits.conf, not in systemd

Oops, I left out `roles/manage_operating_system/tasks/enable_core_dump.yml`, amended and pushed.  This affect the first 2 bullets.

> > * Do not adjust OS default setuid for potential core dumps

This was removed here: https://github.com/EnterpriseDB/edb-ansible/pull/600/files#diff-385a6f45153d7ee74e427aacefdcc208ce959f822314c84d440ca3ab025514d3L36

> > * Stop overriding core dump locations, which confuses distro specific helper tools
> 
> I know for the last item, not overriding the location essentially means not defining our own core dump directory, as we have been doing. However, I don't see any updates to the tasks within the `manage_operating_system` role?

Right, the issue is because different distros manage this differently.  They may use different locations from each other.  On top of that, different distros have their own distro specific tools for managing/analyzing/etc.  The couple that I've looked at do not like it (i.e. can't handle) a different location than their own defaults.  So we have to be aware of what that location is is for whatever distro we're using.

roles/install_dbserver/tasks/EPAS_RedHat_install.yml
```diff
@@ -70,3 +70,15 @@
     state: present
   become: true
   when: pg_ssl
+
```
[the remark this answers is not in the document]
mw2q: Yeah, accidentally left out, but amended and push commit to include. `roles/manage_operating_system/tasks/enable_core_dump.yml`.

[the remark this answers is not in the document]
hannahms: Also, with the new `roles/manage_operating_system/tasks/enable_core_dump.yml`, do you plan on importing these tasks into a benchmark playbook? I think there should still be a way into `enable_core_dump.yml` through the `roles/manage_operating_system/tasks/main.yml` 
```
- name: Enable core dumps
  ansible.builtin.include_tasks: enable_core_dump.yml
  when:
    - enable_core_dump | bool
```
and add the `enable_core_dump: false` var into the `roles/manage_operating_system/defaults/main.yml` to keep consistent with `enable_user_profiling`

roles/manage_operating_system/tasks/enable_core_dump.yml
```diff
@@ -0,0 +1,15 @@
...
+    limit_item: "core"
+    value: "unlimited"
+  become: true
+
+- name: Install debuginfo helper packages
+  ansible.builtin.package:
+    name: yum-utils
+    state: present
+  when: ansible_os_family == 'RedHat'
+  become: true
```
hannahms, line 15: the one thing i have with placing this here is about needing to run this `manage_operating_system` role prior to the `install_dbserver` role, i think the install tasks should be run in the same place

roles/install_dbserver/defaults/main.yml
```diff
@@ -6,7 +6,6 @@ pg_version: 14
 pg_tuner_version: 1
 pg_owner: "{{ 'enterprisedb' if pg_type == 'EPAS' else 'postgres' }}"
 enable_core_dump: false
```
hannahms, line 8: was this intended to be left in place? I noticed they were removed in other roles' defaults

roles/install_dbserver/tasks/EPAS_RedHat_install.yml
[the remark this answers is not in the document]
mw2q: > if the `yum-utils` is installed in the `manage_operating_system` tasks, should the debuginfo packages also be installed in the same place? I don't see the need to break it up

I think this might be a chicken and egg issue.  We can't know what debug packages to install until we know what packages to install, but we need to have the tool to install debug packages before we try.

At least I think it makes sense to install the tool so that we can install debug packages, and the install them from the respective playbooks when it's known what the packages are.

I'm open to other ideas.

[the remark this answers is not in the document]
mw2q: > - name: Enable core dumps
>   ansible.builtin.include_tasks: enable_core_dump.yml
>   when:
>     - enable_core_dump | bool

Whoops, thanks for catching that.  That's what I intended but I think I lost of changes swapping between system.  Amended commit and forced push.

roles/manage_operating_system/tasks/enable_core_dump.yml
mw2q: I think we can move it somewhere else, but it needs a more general location than `install_dbserver` because it can apply to non-database systems.  Do you think an `install_os` role or something is needed? (I don't think install_os is a good name...)

roles/install_dbserver/defaults/main.yml
mw2q: > was this intended to be left in place? I noticed they were removed in other roles' defaults

Yeah, I should I should leave `enable_core_dump` in there because it was used in `roles/install_dbserver/tasks/EPAS_RedHat_install.yml`.  Should I handle that differently?

roles/install_dbserver/tasks/EPAS_RedHat_install.yml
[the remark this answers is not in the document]
hannahms: what if we kept it here but also placed it in the `install_dbserver` role? Like this in `install_dbserver` and keep the current `yum-utils` install in `manage_operating_system` because if it is already installed, this is not expensive to run and will be skipped? 
```
- name: Install debuginfo helper packages when enable_core_dumps
  ansible.builtin.package:
    name: yum-utils
    state: present
  when:
    - enable_core_dumps is defined
    - enable_core_dumps | bool
  become: true
```

roles/manage_operating_system/tasks/enable_core_dump.yml
hannahms: I understand, I think leaving it here is a good place.

hannahms (changes requested):
Add variable `enable_core_dump: false` to variable defaults `roles/manage_operating_system/defaults/main.yml`

roles/install_dbserver/defaults/main.yml
hannahms: no, i think I initially thought this wasn't present, which is why I asked you to include the condition `when: enable_core_dump is defined`, but I see this now. I don't think keeping the conditionals with the `is defined` is an issue

mw2q: Ok, I'll re-edit that `is defined`. ;)

roles/install_dbserver/tasks/EPAS_RedHat_install.yml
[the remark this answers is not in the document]
mw2q: > what if we kept it here but also placed it in the `install_dbserver` role? Like this in `install_dbserver` and keep the current `yum-utils` install in `manage_operating_system` because if it is already installed, this is not expensive to run and will be skipped?

Ok, I can put it back.  I don't see the harm, I'm not sure if it's useful without the other operating system changes, but we can see how it goes.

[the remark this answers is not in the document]
hannahms: i understand, but I think including so that nothing fails here if this is run prior to the `manage_operating_system` role is useful

roles/install_dbserver/defaults/main.yml
hannahms: you can leave as is if you would like to edit less files :)

mw2q:
Make another pass and forced push.

mw2q:
> Add variable `enable_core_dump: false` to variable defaults `roles/manage_operating_system/defaults/main.yml`

I think i missed this earlier, but got it now.

hannahms (approved):
LGTM

PR #602, opened 2023-10-11
install fonts for appimages

mw2q:
gnuplot in the AppImages rely on fontconfig and some fonts to be installed on the system.


hannahms (approved):
LGTM

PR #603, opened 2023-10-17
Update dbt3 roles to use AppImage

mw2q:


roles/setup_dbt3/defaults/main.yml
hannahms, line 1: If I'm not mistaken, this role no longer requires the `install_dbserver` and `init_dbserver`, where the var `pg_user_home` was set. I think the var `pg_user_home` should be defined in the `roles/setup_dbt3/vars/*db  type*.yml`, it is used in the `dbt3_tpc.yml` tasks and `dbt3_configure_profile.yml`

roles/setup_dbt3/defaults/main.yml
```diff
@@ -1,10 +1,14 @@
 ---
-dbt3_version: "1.11.0"
+dbt3_version: "2.0.1"
 dbt3_pgdata: "/pgdata/dbt3/pgdata"
-pg_version: 14
```
hannahms, line 4: If `pg_version` is removed here, I think the variables within the `roles/setup_dbt3/vars/**` files should be reviewed, some unused vars (`pg_data`) use `pg_version`, not sure if this will throw errors while importing the vars but still a good idea to clean up

hannahms (changes requested):
Great work on the AppImage! Just a few minor details to double check variable definition

roles/setup_dbt3/defaults/main.yml
mw2q: Ah, I don't think the `vars` stuff is needed at all now.  Removing.

roles/setup_dbt3/defaults/main.yml
mw2q: Right, I'll add `pg_user_home`...

mw2q:
ok, forced push updates

hannahms (approved):
LGTM

PR #608, opened 2023-10-26
Added support for Ubuntu22.

bonesmoses:
Given that edb-ansible supports Ubuntu20, Ubuntu22 should also be compatible. I didn't physically test every single role, but have worked extensively with install_dbserver, setup_patroni, and setup_etcd, which are all necessary to set up a basic Patroni cluster. None of these were affected simply by adding "Ubuntu22" as a patch, which was necessary for various unrelated tests.

### DIFF
--- a/roles/execute_binary_upgrade/defaults/main.yml
+++ b/roles/execute_binary_upgrade/defaults/main.yml
@@ -52,6 +52,7 @@ supported_os:
   - RedHat8
   - RedHat9
   - Ubuntu20
+  - Ubuntu22
   - Debian9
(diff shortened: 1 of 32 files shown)


bonesmoses:
I took a look at the failed checks, and they were related to not being able to pull edb-ansible from ansible-galaxy, which is... odd.

hannahms (approved):
thank you for this contribution!

hannahms:
galaxy has a new protocol, this issue has been sporadically popping up, so nothing too alarming

PR #616, opened 2023-12-13
force pgbackrest template config file name

aya:
When using the same server for both primary and pgbackrestserver roles, pgbackrest role will fail with following error :
```
Could not find or access
'./templates/pgbackrest_pgbackrestserver.async.conf.template'
```
This is due to server belonging to both 'pgbackrestserver' and 'primary' inventory groups.


vibhorkumar123 (approved):
LGTM

PR #617, opened 2023-12-14
add ability to override tuning calculated values

aya:
On servers with low ram or custom footprint, it may be useful to manualy set specific config values like shared_memory and let the tuning role set other values automaticaly.


vibhorkumar123 (changes requested):
The parameter name is very long. Please use parameter like
pg_conf_override_tuning

aya:
I updated the parameter name and rebased onto main branch.

vibhorkumar123:
LGTM. Except ansible-lint is failing. Please look into it.

PR #618, opened 2024-02-19
Fix yum.postgresql.org repository GPG key path

Ianvdl:
The current path for the PostgreSQL yum repository GPG key is invalid, and returns a 404 error: http://yum.postgresql.org/RPM-GPG-KEY-PGDG

The key is in fact at: http://yum.postgresql.org/keys/RPM-GPG-KEY-PGDG

This patch fixes that.


Ianvdl:
I see there's a couple more occurences elsewhere, I'll add those to the PR first.

vibhorkumar123 (approved):

vibhorkumar123:
LGTM!

PR #624, opened 2024-04-02
Adjust python-apt package to be findable on newer Debian/Ubuntu versions

Ianvdl:
At the moment `python-apt` is installed on Debian-based distros using the following:

```
 - name: Install python-apt package
   ansible.builtin.package:
     name: python-apt
     state: present
   when:
     - ansible_os_family == 'Debian'
```

On newer versions of these distros, the `python-apt` package has been superseded by `python3-apt`, so the playbook fails at this point. Since all other references in `roles/install_dbserver/tasks/validate_install_dbserver.yml` are already configured to use the `python3` versions, e.g.:

```
grep python3 roles/install_dbserver/tasks/validate_install_dbserver.yml
        'python3-pycurl', 'python3-libselinux', 'python3-psycopg2', 'glibc-langpack-en'
        'python3-pip', 'python3-psycopg2'
        'ca-certificates', 'python3-pycurl', 'python3-psycopg2', 'postgresql-' + pg_version | string,
        'python3-pip', 'python3-psycopg2', 'edb-as' + pg_version | string + '-server',
```

We are probably safe to proceed with this change.

### DIFF
--- a/roles/install_dbserver/tasks/validate_install_dbserver.yml
+++ b/roles/install_dbserver/tasks/validate_install_dbserver.yml
@@ -97,12 +97,21 @@
     - ansible_os_family == 'RedHat'
     - pg_version|int > 10
 
-- name: Install python-apt package
+- name: Install python-apt package for Debian 9/10
+  ansible.builtin.package:
+    name: python-apt
+    state: present
+  when:
+    - ansible_os_family == 'Debian'
+    - ansible_distribution_major_version in ['9', '10']
+
+- name: Install python3-apt package for Debian 11/12
   ansible.builtin.package:
     name: python3-apt
     state: present
   when:
     - ansible_os_family == 'Debian'
+    - ansible_distribution_major_version in ['11', '12']
 
 - name: Set package list for PG Debian
   ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


roles/install_dbserver/tasks/validate_install_dbserver.yml
```diff
@@ -99,7 +99,7 @@
 
 - name: Install python-apt package
   ansible.builtin.package:
-    name: python-apt
```
vibhorkumar123, line 102: I am worried about what would happen to the old version of Debian. Would you like to add a check for the Debian version and, based on that, use python3-apt?

vibhorkumar123 (changes requested):

roles/install_dbserver/tasks/validate_install_dbserver.yml
Ianvdl: I could do that - do you think it's necessary? Debian removed python-apt five years ago, as far as I can tell, along with Python 2.7.

Ianvdl:
Updated the PR to use `python-apt` for Debian 9 and 10 (Stretch and Buster), and `python3-apt` for Debian 11 and 12 (Bullseye and Bookworm).

vibhorkumar123 (approved):
LGTM!

PR #626, opened 2024-04-09
Disable RPM GPG key checks for PGDG repo and fix key paths

Ianvdl:
This should allow the Github tests to start succeeding as well. It seems that the GPG keys are for the packages in the yum repo, not for the RPM package to install the repo, so it doesn't succeed with a GPG check.


vibhorkum (approved):
LGTM!

vibhorkumar123 (approved):

PR #627, opened 2024-04-10
Fix linting issues due to long lines and improper spacing

Ianvdl:
These were introduced at #617 and prevents all checks from succeeding.


vibhorkumar123 (approved):

Ianvdl:
I see there's one more failing elsewhere, following the check above.

Ianvdl:
Okay, last one fixed, it should succeed now.

vibhorkumar123 (commented):
LGTM!

PR #629, opened 2024-04-15
Allow setting EFM detach.on.agent.failure property.

Ianvdl:
Keeping the default as true.


vibhorkumar123:
It's showing come conflicts. Can you please look into it?

Ianvdl:
> It's showing come conflicts. Can you please look into it?

Should be fixed now - it's because of the other EFM merge :)

PR #631, opened 2024-04-18
Remove jinja2 templating from assertion predicates

Ianvdl:
This addresses a fatal error in newer Ansible versions caused by a warning stating that "conditional statements should not include jinja2 templating". This warning leads to the fatal error stating that the conditional check is marked as unsafe.

This patch removes the jinja templating from the strings and also removes the strings themselves, leaving the bare predicate.


vibhorkumar123 (approved):
LGTM

PR #632, opened 2024-04-25
Make PgPool PCP package name configurable on RHEL

Ianvdl:
At the moment, the PgPool package can be specified manually, but the corresponding pcp package on RHEL cannot. This patch allows for that, keeping the existing defaults. A patch is not needed for EPAS, as it does not make use of such a package.


vibhorkumar123 (approved):
LGTM!

PR #633, opened 2024-04-25
Add run_once to PgPool test user creation

Ianvdl:
In some cases I have found that the task to create the PgPool test user can run multiple times, with runs subsequent from the first failing with a duplicate key error (the user already exists on the backend after the first run). This patch prevents that from occurring.


vibhorkumar123 (approved):
LGTM!

PR #634, opened 2024-04-25
Add minimum EFM PgPool integration values

Ianvdl:
At the moment there is an option to enable EFM PgPool integration, but the integration is incomplete as it does not include the minimum parameters required by EFM to complete the integration. As such, the flag is currently not viable, even though it exists.

This patch fixes this by making use of existing variables to populate the configuration when the flag is enabled, and allows for the attach and detach scripts to be made configurable.


vibhorkumar123 (approved):
LGTM

